PR #117, opened 2014-06-16
use Lwt_log everywhere

hhugo:


src/lib/eliom_lib.server.ml
```diff
@@ -14,8 +14,6 @@ include (Eliom_lib_base : module type of Eliom_lib_base
 
 let escaped_value_escaped_value = fst
 
-let debug f = Printf.ksprintf (fun s -> Printf.eprintf "%s\n%!" s) f
-
```
Drup: @hhugo You reverted this change in the client, but not the server (and I missed it >_>).

PR #157, opened 2014-11-22
Add services for HEAD, OPTIONS and PATCH methods.

slegrand45:
Useful for REST services: http://restful-api-design.readthedocs.org/en/latest/methods.html


vasilisp:
Closing old PRs. `Eliom_service` has changed a lot, so I don't know how usable that one could be. Thanks anyway!

Issue #166, opened 2015-01-30
eliomdep returns wrong exit codes


pveber:
eliomdep returns non-zero exit codes when invoked, although it worked well, as illustrated in the following example:
```
$eliom-distillery -name delme
Destination directory "delme" doesn't exists. Create it? (YES/no) 
Generated delme/static/css/delme.css.
Generated delme/README.
Generated delme/Makefile.options.
Generated delme/Makefile.
Generated delme/delme.conf.in.
Generated delme/delme.eliom.
$cd delme/
$eliomdep -server -sort  delme.eliom > rien
$echo $?
118
$eliomdep -server -sort  delme.eliom > rien
$echo $?
126
$eliomdep -server -sort  delme.eliom > rien
$echo $?
134
```
The error is that when invoked with `-sort`, `eliomdep` eventually returns with the exit code provided by the function `sort` in `eliomdep.ml`, which itself returns the integer provided by the function `create_process` from `utils.ml`. This function relies on `Unix.create_process` which **does not** return the exit status of the invoked shell command, but its PID (which explains why `eliomdep` returns a different integer at each call). 

The fix should probably be to modify the `create_process` function, but as it may have other implications, I refrain from proposing a patch until someone confirms this is an adequate way out of this bug.

PR #167, opened 2015-01-30
eliomdep: fixed exit code when calling option "-sort"

pveber:
Details of the problem are given in issue #166. I found a minimal fix that shouldn't impact any other function.

### DIFF
--- a/src/tools/eliomdep.ml
+++ b/src/tools/eliomdep.ml
@@ -183,7 +183,8 @@ let sort () =
   in
   create_process !compiler
     ( "-sort" :: "-pp" :: get_pp pkg ppopt :: eliom_synonyms @
-        List.(concat (map (fun file -> ["-impl"; file]) !sort_files)) )
+      List.(concat (map (fun file -> ["-impl"; file]) !sort_files)) )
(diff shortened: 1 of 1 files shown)


Drup:
Thanks!

pveber:
Woops, looks I shouldn't have used `|>`, which was not available in 4.00.1 (see [Jenkins output](https://ci.inria.fr/ocsigen/job/Eliom-4.00.1/804/changes)). Should I send another patch?

Drup:
Yes please, I forgot to fix it yesterday.

Drup:
In fact, I was going to fix it by modifying Ocsigen_lib instead, in order to add |>.

pveber:
Ok, then it's simpler if I let you do it, not sure where it should go.
Reminder, you can use the "primitive" implementation, for best performance:

```ocaml
external ( |> ) : 'a -> ('a -> 'b) -> 'b = "%revapply"
```

2015-02-03 8:22 GMT+01:00 Gabriel Radanne <notifications@github.com>:

Drup:
I precisely cannot use that if I want retro-compatibility ;)

PR #172, opened 2015-03-01
Makes server function return types covariant.

Drup:
It seems safe and should accept slightly more programs.

In particular, it helps with html subtyping.


Drup:
cc @hhugo : you know this part quite well, an opinion ?

Drup:
ping ? :)

hhugo:
I don't feel like I know this part quite well but I would say go for it.

PR #177, opened 2015-04-30
Fix bug #168 (call_ocaml_service always sends boolean true).

### LINKED ISSUE #168: Eliom_client.call_ocaml_service always send boolean true
The following code (modified version of the pi tutorial), always print `receive true` regardless of the parameter given to `Eliom_client.call_ocaml_service` here `false`. I expect this code to alert with the message `receive false`.

``` ocaml
{shared{
  open Eliom_content
  open Html5
  open Html5.D
}}

module Ex_app =
  Eliom_registration.App (struct let application_name = "graffiti" end)

let pi_service =
  Eliom_registration.Ocaml.register_post_coservice'
    ~post_params:(Eliom_parameter.bool "test")
    (fun () b -> Lwt.return (if b then "receive true" else "receive false"))

let _ =
  Ex_app.register_service
    ~path:["pi"]
    ~get_params:Eliom_parameter.unit
    (fun () () ->
      ignore {unit{
        Lwt.ignore_result(
          lwt s =
            Eliom_client.call_ocaml_service ~service:%pi_service () false
          in
	  Lwt.return (
	    Dom_html.window##alert(Js.string ("answer = "^s))))
      }};
      Lwt.return
        Html5.D.(html
                  (head (title (pcdata "pi")) [])
                  (body [])))
```

mfp:
Fix #168.

### DIFF
--- a/src/lib/eliom_parameter.server.ml
+++ b/src/lib/eliom_parameter.server.ml
@@ -260,8 +260,10 @@ let reconstruct_params_
         | TAtom (name,TBool) ->
           (try
              let v,l = (List.assoc_remove (pref^name^suff) params) in
-             Res_ (true,l,files)
-           with Not_found -> Res_ (false, params, files))
+             Res_ (atom_of_string TBool v,l,files)
+           with
+             | Not_found -> Res_ (false, params, files)
+             | e -> Errors_ ([pref^name^suff,"",e], [], files))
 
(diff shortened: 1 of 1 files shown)


mfp:
I think the build was broken before this, and the patch is OK (it does build when applied atop 4.1 against ocsigenserver 2.5).

PR #186, opened 2015-07-23
fixing encoding of Boolean parameters (true -> "on", false -> nothing)

vasilisp:
This patch fixes how Boolean parameters are encoded as part of the URL. Our convention is that any string for the parameter means true, and absence of the parameter means false.


vasilisp:
I will push the same fix to master once this gets merged.

PR #187, opened 2015-08-06
Update for new-style TyXML wrapping

vasilisp:
This brings Eliom up-to-speed with the new-style wrapping available in current TyXML master. See ocsigen/tyxml#58 for the TyXML side of the story. This PR presupposes ocsigen/js_of_ocaml#348.

### DIFF
--- a/.jenkins.sh
+++ b/.jenkins.sh
@@ -1,8 +1,8 @@
 opam pin add --no-action eliom .
 opam pin add --no-action reactiveData 'https://github.com/hhugo/reactiveData.git#master'
-#opam pin add --no-action tyxml 'https://github.com/ocsigen/tyxml.git#master'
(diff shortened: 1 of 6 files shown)


Drup:
You missed some functor application, it seems. :+1: otherwise.

vasilisp:
Are you referring to the Jenkins output? That's because of JSOO. The JSOO master doesn't compile against TyXML master, and I didn't want to pin a very ephemeral branch.

vasilisp:
A slightly annoying side-effect is that we now get error messages with types like `string Eliom_content.Html5.F.wrap`, which is unnecessarily complicated. Maybe TyXML could provide versions of the signature functors that don't expose wrap and such?

Drup:
That was already the case before. I prefer to tell everyone to use `-short-path`.

PR #188, opened 2015-08-07
Usable reactive class attributes

vasilisp:
This is an effort to make reactive class attributes usable. It contains two patches.

First, the code that tries to merge multiple class attributes into a single attribute is updated to take into account the default values of client-side attributes. If this does not happen, given that Eliom introduces auxiliary classes, we end up with multiple class attributes. Both Firefox and Chromium refuse to render the page. I guess this is is a special case of ocsigen/tyxml#69.

Second, we try to be smart when reactively updating the class attribute. The semantics are as follows. If at a particular point in time the list of classes in the respective attribute in the DOM is `l1`, and the signal goes from `l2` to `l3`, then the list of classes becomes `l3 ∪ (l1 \ l2)`. We add any new classes, we remove the classes that disappear from the reactive signal, and we retain the classes that are neither in the old nor in the new value of the signal, i.e., the "non-reactive" classes. Does this sound reasonable?

### DIFF
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -514,6 +514,25 @@ let rebuild_attrib_val = function
   | Xml.AStrL (Xml.Space, sl) -> Js.string (String.concat " " sl)
   | Xml.AStrL (Xml.Comma, sl) -> Js.string (String.concat "," sl)
 
+let class_list_of_racontent_o = function
+  | Some (Xml.AStr s) ->
+    [s]
+  | Some (Xml.AStrL (space, l)) ->
+    l
+  | Some _ ->
+    failwith "attribute class is not a string"
+  | None ->
+    []
+
+let rebuild_class_list l1 l2 l3 =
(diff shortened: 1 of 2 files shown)


Drup:
Do you think we could implement a more general version of 1) by adding a pass that makes attributes uniques inside the `Xml.leaf` and `Xml.node` combinators ?

vasilisp:
I think it's quite complicated.

We don't really know what to do with the duplicates. For `class`, it is somewhat meaningful to accumulate the contents, but for other attributes not so much (e.g., only one ID is allowed). It has to be attribute-specific. We could perhaps default to keeping just one instance, with special treatment for class.

Also, the patch targets client attributes, which are specific to Eliom. These don't necessarily have a known value (or even an attribute name) at initialization time, and they largely bypass TyXML.

Drup:
Fair enough, the code looks fine to me. I'll let vincent be judge of the semantic, since he used reactive node more.

PR #189, opened 2015-08-21
Syntax for Shared Expressions

vasilisp:
This PR introduces syntax for shared expressions. The syntax is as follows:

    {shared# t { e }}

where `t` is a type, and `e` is an expression. `t` is optional.

I initially tried `|` instead of `#`, but that was causing trouble for Tuareg / ocp-indent. The above syntax is indented in a fairly decent way.

Allowing `{shared{ e }}` would introduce ambiguity: the same syntax is used for `str_item`, and expressions are allowed in the context of `str_item`.

We are transitioning to PPX, but the transition is not happening overnight, and it is good to have more usable shared expressions in the meantime.

### DIFF
--- a/src/lib/eliom_lib.client.ml
+++ b/src/lib/eliom_lib.client.ml
@@ -35,6 +35,9 @@ exception Exception_on_server of string
 
 type 'a client_value = 'a
 type 'a shared_value = 'a
+
+let create_shared_value (_ : 'a) (c : 'a client_value) = c
+
 (*****************************************************************************)
 
 module Url = struct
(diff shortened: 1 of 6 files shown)


hhugo:
The main reason not to use `{shared|` is not tuareg/ocp-indent, but rather that this syntax is already used by the ocaml language for strings.

Drup:
Do we really need the sharp ?

Why not just `{shared t { ... }}` ?

vasilisp:
@drup, we do not strictly need the sharp. It is just that `{shared t { ... }}` is currently valid syntax for client expressions of type `shared t`. But I have no strong feelings, we can omit the sharp, or use another character.

PR #197, opened 2015-10-15
Eliom_client.unload: ability to ask for confirmation before leaving

vasilisp:
This PR enhances the `onunload` callback handling to allow asking for a confirmation. If some callback returns `Some s` (for a string `s`), then a confirmation message *usually* showing `s` appears. Note that the signature changes.

We react both to the `onbeforeunload` event, and to page changes initiated by calling a service. We try to provide consistent behavior in the two cases. The string `s` is not used by Firefox in the `onbeforeunload` case ([FF bug 578828](https://bugzilla.mozilla.org/show_bug.cgi?id=578828)).

Multiple callbacks are allowed, and we try to execute the remaining ones if the user decides to stay. This happens via an `onunload` handler if the whole thing has started by `onbeforeunload`.

The custom `addEventListener` wrapper (`add_string_event_listener`) is because WebKit cares about the return value of the callback, so we need to return a string and not a Boolean. I removed the check for `(Js.Unsafe.coerce o)##addEventListener == Js.undefined` because `Dom.addEventListener` and `add_string_event_listener` take care of this.

### DIFF
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -79,23 +79,59 @@ let int64_to_string i = (Obj.magic i)##toString()
 
 let create_buffer () =
   let elts = ref [] in
-  let add x =
-    elts := x :: !elts
+  let add x = elts := x :: !elts
+  and get () =
+    (* NOTE: the caller needs to reverse the list *)
+    !elts
   in
   let flush () =
-    let res = List.rev !elts in
+    let res = List.rev (get ()) in
     elts := [];
     res
   in
-  add, flush
+  add, get, flush
 
 (* == Callbacks for onload and onunload *)
 
 let run_callbacks handlers = List.iter (fun f -> f ()) handlers
 
-let onload, flush_onload = create_buffer ()
+let onload, _, flush_onload = create_buffer ()
 
-let onunload, flush_onunload = create_buffer ()
+exception Cancel_onunload of string
+
+let onunload, run_onunload =
+  let add, get, flush = create_buffer () in
+  let r = ref (get ()) in
+  let add x = add x; r := get () in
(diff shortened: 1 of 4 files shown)


src/lib/eliom_lib.client.ml
```diff
@@ -99,8 +99,14 @@ let lwt_ignore ?(message="") t =
 let jsalert a = Dom_html.window##alert (a)
 let alert fmt = Printf.ksprintf (fun s -> jsalert (Js.string s)) fmt
 
-let debug_var s v = Js.Unsafe.set Dom_html.window (Js.string s) v
+let confirm =
+  let f s =
+    let s = Js.string s in
+    Dom_html.window##confirm(s) |> Js.to_bool
+  in
+  fun fmt -> Printf.ksprintf f fmt
```
Drup: We should move all those functions to `Format`, it composes, better, but it's orthogonal to the current patch.

src/lib/eliom_client.client.ml
```diff
@@ -80,23 +80,55 @@ let int64_to_string i = (Obj.magic i)##toString()
...
+  let add x = elts := x :: !elts
+  and get () = List.rev !elts in
   let flush () =
-    let res = List.rev !elts in
+    let res = get () in
     elts := [];
     res
   in
-  add, flush
+  add, get, flush
```
Drup: I feel we should just use `Queue`, from the stdlib, instead, but it's not very important.

Drup:
I completely misunderstood when you explained me what this patch was about a few days ago. This API is fine. I don't have further comment on the main piece of the patch, I trust you tested it enough.

> The string s is not used by Firefox in the onbeforeunload case (FF bug 578828).

Please add that to the documentation.

PR #202, opened 2015-10-19
Use relative URLs for client-server communication

vouillon:


src/lib/eliom_request.client.ml
```diff
@@ -199,6 +199,7 @@ let send ?(expecting_process_page = false) ?cookies_info
       else true
     in
     try_lwt
+Format.eprintf "XXXX %s@." (Url.encode_arguments get_args);
```
Drup: Hum ?

Drup:
Why where we using absolute URLs before ?

vasilisp:
Does this fix any issues (with redirects maybe) when behind a proxy?

I have no comments on the code.

vouillon:
This make Eliom work behing a proxy that does TLS termination. Otherwise, the server returns https URLs instead of http URLs. Some proxys set a ```X-Forwarded-Proto``` header, which is recognized by the accesscontrol Ocsigen extension. But Cloudfront, for instance, sets ```Cloudfront-Forwarded-Proto``` instead...

This also makes it possible to rewrite URLs between the server and the client without having Eliom aware of that.

Drup:
I see. :+1: then :)

Issue #200, opened 2015-10-16
Deprecate pa_include and remove the sigs directory


vasilisp:
Apologies if this has been discussed before.

I wonder whether we really need `pa_include` and the `sigs` directory. We can define signatures and use standard features of the module language to include them. Am I missing something?

`sigs` is not even properly integrated into the build system.

Now (during the transition to PPX) is the right time to do it. We will be better off without a `ppx_include` preprocessor. Objections?

vasilisp:
I just realized that `ppx_deriving` already exists, but still, do we need to use it?

Drup:
The reason `pa_include` exists is because some substitution are not authorized in OCaml signatures, in particular when type arity doesn't match. We removed some of them over the year, so maybe it's okay to clean up that mess now. 

At least we can remove as many usages as possible.

vasilisp:
Progress by #293.

Issue #199, opened 2015-10-16
Use Raw forms by default?


balat:
By default Eliom is not using tyxml interface for forms but the interface compatible with the service identification mechanism (Eliom_parameters and so on).

But:
 - it makes difficult to understand the API doc (we don't understand why it is not like in tyxml)
 - Client server apps do not use this interface, forcing to use Raw module for each form element.

I think it would be cleaner to keep the usual tyxml interface by default, and open a module for Eliom_parameter-like form elements.

But it will break existing code :/
What do you think?

Drup:
I agree with the issue. It becomes especially annoying when you want to have some javascript treatment (like validating a field) before sending it.

I think it would be a good occasion to rework the form interface in general to be more flexible. Propose the interface into a new module/package `Eliom_form`, ask people to migrate, then remove the special things a version later, or something like that.

vasilisp:
It seems inconsistent to me that the `D` and `F` modules don't respect the standard TyXML signatures. In fact, I am not happy with the very existence of the `Raw` modules :). So in principle I agree with the change.

vasilisp:
In fact, the separate module that @Drup proposes sort-of exists. `Eliom_registration_base` has all the form stuff, and little connection to `Eliom_registration`. So we can fairly easily clean-up the namespace (and `ocamldoc` output, etc.) by renaming `Eliom_registration_base` and making it public. Thoughts?

Should that change be in the release? It doesn't seem to be a lot of work.

Drup:
`_base` also doesn't have overloaded `a` (and various other things, iirc).

vasilisp:
Closed by #215.

PR #206, opened 2015-10-29
Clean-up Forms

vasilisp:
This PR restructures the forms support in Eliom.

`Eliom_form_sigs` replaces the form-related signatures under `src/lib/sigs`. This is a step towards addressing #200.

`Eliom_registration_base` becomes `Eliom_form`. I functorized the interface to remove the dependency on `Eliom_content_core`. Internally, `Eliom_form` is much simpler than the venerable `Eliom_registration_base` (because we are now dealing with homogeneous form-related signatures).

The `Eliom_content.{F,D}` signatures (both server- and client-side) refer to `Eliom_content_sigs.FORMS`. To ensure that the actual API has not changed, I didn't modify the `pa_include`-enabled signature until 98c3a2631b7e537882882114e2740d5242502022. So if there is API breakage, one of the last two commits is to blame.

Clearly, more work is needed on forms (to solve #199 among other issues). This is about where the non-breaking changes end. It would be nice to get this merged, so that the form-related work can go on without diverging too much from master :).


src/lib/eliom_content_.client.ml
```diff
@@ -396,7 +407,18 @@ module Html5 = struct
...
+      Eliom_form.Make(struct
+        include Html5.D
+        module Svg = Eliom_content_core.Svg.D
+        let uri_of_fun = Eliom_content_core.Xml.uri_of_fun
+        let attrib_of_service s info =
+          Eliom_content_core.
+            (Html5.D.to_attrib
+               (Xml.internal_event_handler_attrib
+                  s (Xml.internal_event_handler_of_service info)))
+        let attrib_onclick = Eliom_content_core.Html5.D.a_onclick
```
Drup: `attrib_onclick` seems to always be the same than `a_onlick`. Maybe it could be removed ?

src/lib/eliom_form_sigs.shared.mli
```diff
@@ -0,0 +1,1034 @@
...
+module type PARAM = sig
+
+  type +'a elt
+  type +'a attrib
+
+  type uri
+
+  val a_input_required : unit -> Html5_types.input_attrib attrib
+
+  val a_select_required : unit -> Html5_types.select_attrib attrib
```
Drup: If we follow https://github.com/ocsigen/tyxml/pull/51, remove the argument completely and make it a constant instead.

src/lib/eliom_form_sigs.shared.mli
```diff
@@ -0,0 +1,1034 @@
...
+  | `Week
+  | `File
+  | `Email
+  | `Image
+  | `Datetime_local
+  | `Datetime
+  | `Date
+  | `Color
+  | `Button
+  ]
```
Drup: Isn't this the same than `Html5_types.input_type` ? Also, it doesn't seem to be used.

src/lib/eliom_form_sigs.shared.mli
```diff
@@ -0,0 +1,1034 @@
...
+
+type button_type =
+  [ `Button
+  | `Reset
+  | `Submit
+  ]
```
Drup: See above comment.

src/lib/eliom_form_sigs.shared.mli
```diff
@@ -0,0 +1,1034 @@
...
+    [> Html5_types.script] elt
+
+  val make_for_attrib : string -> [> `For] attrib
+
+end
+
+module type S = sig
+
+  (* When modifying this interface, please ensure that the ocamldoc is
+   coherent with the ocamldoc from Eliom_uri. *)
```
Drup: Split the signature and use include ? It's normal include, so it should be linkified properly in ocamldoc.

src/lib/eliom_form.eliomi
```diff
@@ -0,0 +1,60 @@
+(* Ocsigen
+ * http://www.ocsigen.org
+ * Module Eliom_predefmod
```
Drup: Nope. ;)

Drup:
Very nice work. I was going to make more comments but it was about breaking the interface, so it will be for the next set of changes. 

If my understanding is correct. `Eliom_form` is the base wiring and `Eliom_mk_form` add the typed versions ? I think we should change the name or even just merge them.

src/lib/eliom_form.eliom
```diff
@@ -0,0 +1,337 @@
...
+    ?a: (('a attrib) list) ->
+    ('b elt) list Eliom_lazy.request ->
+    'c elt
+
+  val lazy_form:
+    ([< Html5_types.form_attrib ],
+     [< Html5_types.form_content_fun ],
+     [> Html5_types.form ]) lazy_star
+
+  val uri_of_fun : (unit -> string) -> Xml.uri
```
Drup: I know you need it for compat but, hum, why aren't we using `Xml.uri_of_string` here ?

Drup:
After taking another look at the functors in Eliom_(mk)form ... why so many ? Can you give a decent explanation of the separation ? At the end, you only expose the composition of 3 functors, which makes me thing the separation is not that useful.

src/lib/eliom_form_sigs.shared.mli
Drup: And if we don't (it's not merged in tyxml yet ..). Just put the `` `Required`` argument back

vasilisp:
High-level points first: I agree with you that the separation between `Eliom_(mk)form(s)` and the different levels of functors are not that useful. I just didn't get around to removing all the separation that existed before.

Right now, `Html5_forms_base` produces the basic combinators that `Eliom_mkforms.MakeForms` uses, `Eliom_mkforms` adds the typed versions as you say, and `MakeApplForms` adds some extra stuff. I am pretty sure we can collapse the last two levels, and probably all three.

I added the top-level `Make` to hide all the plumbing currently needed, and it will go away when we collapse things.

I don't see any reason for retaining `Eliom_mkforms` as a separate module.

I would rather do the above transformations in the process of reducing what is now `Eliom_mkforms` to its essence (via internal functorization, or via GADTs; we will discuss the interface and associated implementation).

Drup:
Ok, that sounds good to me.

src/lib/eliom_form_sigs.shared.mli
vasilisp: `PARAM` is an internal signature that will go away when we collapse things, so it doesn't really matter :).

src/lib/eliom_form_sigs.shared.mli
vasilisp: Good catch ;), I just de-inlined it without noticing. It is used (`Eliom_content_sigs`).

Issue #196, opened 2015-10-14
shared expressions inside shared expressions


balat:
It makes sense to allow shared expressions inside shared expressions. See the example below.
Would it be easy to allow that?

```ocaml
{shared{
  let mk_selected i =  
    Eliom_shared.React.S.map {shared#{ fun sel -> if ... then ["selected"] else [] }} selected_signal
  in
  Eliom_shared.React.S.map {shared#{
    fun v -> D.div ~a:[R.a_class (%mk_selected i)] [...]
  }}
  signal
}}
```

vasilisp:
I wouldn't call it easy, but maybe it is doable :).

One issue that immediately comes up is that we get unbounded levels of nesting (shared expression inside shared expression inside ...). This is not the case currently, so we need to implement a stack of contexts.

Also, we have to duplicate (or intelligently share) client closures arising from what is textually the same code. We need to think what happens with side effects, with injections referring to different things, etc. (There is already duplication for client expressions appearing in shared expressions, but the problem becomes even more apparent now.)

balat:
Actually the code I wrote does not work. It produce a runtime error "cannot wrap functional values" when called from server side.
I'm still looking for a workaround.

vasilisp:
On the server, your `mk_class` is a function: there is a `fun sel ->` outside the `shared` section. So `mk_class` cannot be injected.

To be honest, I do not understand what you are trying to achieve :). Assuming `mk_selected` and `mk_class` are meant for the same thing, `v` would have to be a shared signal (since it is passed to `mk_class`), but then you apply `Eliom_shared.React.S.map {shared#{ fun v -> ... }} signal`, so `signal` would need to be of type `bool Eliom_shared.React.S.t Eliom_shared.React.S.t`, and I doubt that's what you want. Are `v` and `signal` meant for different things?

balat:
Yes sorry mk_class and mk_selected are the same. I did not simplify the code correctly. I fixed the code (hopefully) above on github.

The code does not work from server side and it's normal.
But how to make it work? Is it possible to do that with the current implementation?
It's not very urgent in my case (I changed the reactive class into a effectful signal changing classes)
but I found not easy solution.

Basically I have a signal called signal (a list of items in my case).
Everytime it changes I want to update the page (display the list).
And inside this part, I want a reactive class (depending on another signal indicating which item is selected and the item number).

This would give something like:
```
Eliom_shared.React.S.map {shared#{
    fun l -> 
... 
[...]

PR #212, opened 2015-11-06
Eliom ppx

Drup:
It's starting to get into a decent shape, so here it is. Documentation is included.

I used the paper for documentation, instead of the camlp4 documentation, since I think it's much simpler and straight to the point. Maybe I will backport that.

Two remaining points
- test on big examples (I only did small things)
- Figure out whether we need the nested_* stuff.

@hhugo: please review the build/packaging part.


Drup:
[Link to the documentation](https://github.com/ocsigen/eliom/blob/ppx/src/ppx/ppx_eliom.mli)

pkg/filelist.ml
```diff
@@ -139,6 +139,12 @@ let ocamlbuild = {
 
 }
 
+let ppx = {
+  interface = [ "ppx_eliom" ; "ppx_eliom_client" ; "ppx_eliom_type" ; "ppx_eliom_server" ];
+  internal = [ "ppx_eliom_utils" ];
+}
```
vasilisp: Add an `interface_only` field here. This is for `mli`-only modules, which have started appearing in Eliom. This fixes the build failure

src/lib/eliom_pervasives.client.ml
```diff
@@ -1,6 +1,8 @@
 
 type 'a client_value = 'a Eliom_lib.client_value
 
+type 'a fragment = 'a client_value
```
vasilisp: Do we really need a new name? "Fragment" could just as well indicate an `str_item`.

vasilisp:
I don't think the `nested_*` stuff is needed, since you expand out shared expressions. At least for now; if we allow for arbitrary nesting as per #196, the need for separation may re-arise.

Drup:
I'm pretty sure the nesting already works, actually.

src/lib/eliom_pervasives.client.ml
Drup: I intend to progressively change the name everywhere over the next versions.

vasilisp:
`js_of_eliom` tries to pipe the shared file `eba_group.mli` through the client preprocessor. As a result, the interface ends up almost empty and EBA fails to compile.

Drup:
@vasilisp I'm wondering how that could work with camlp4. It's basically the same process.

vasilisp:
@Drup, the process was subtly different :). 832bc0dec3bd6286fc606080c5bb283df360460f fixes the issue for me.

Drup:
I see, I still don't really get why it works in camlp4.

vasilisp:
I can't figure out how to use `eliomc` et al. on code that uses other PPX extensions. Preprocessing happens via the `ppx_eliom_*` binaries which (as far as I can tell) do not allow loading other extensions.

vasilisp:
You can find a few more patches in the `ppx-fix-eliomc` branch (for now it is quick fixes, I will clean things up later).

Also, there is an issue with type inference. Given injections `~%bar` and `~%Foo.bar`, `Foo.bar` gets the same type as `bar`.

Drup:
The name collision issue should be fixed.

vasilisp:
With some `sed` preprocessing (coupled with b3b5561609c162b344a361151b6c9217be09d567) to deal with non-parsable `type_mli` files, and with a hackish replacement for `eliomdep -sort` (needed to handle a mix of Camlp4 and PPX files), I have managed to get our mid-sized Camlp4-to-PPX "case study" to compile. Now there are issues with missing injections and closures on the client, and I have no clue what is going on.

Drup:
Can you show me the piece of code involved ?

Drup:
Apparently, it's all fixed now.

src/ppx/ppx_eliom_utils.ml
```diff
@@ -0,0 +1,524 @@
...
+        dispatch_str !context mapper pstr
+    in
+    let loc = {(Location.in_file !Location.input_name) with loc_ghost = true} in
+    open_eliom_pervasives :: Pass.prelude loc @ flatmap f structs @ Pass.postlude loc
+
+  let toplevel_signature context mapper sigs =
+    let f psig =
+      let loc = psig.psig_loc in
+      match psig.psig_desc with
+      | Psig_extension (({txt=("shared"|"client"|"server" as txt)}, PStr strs), _) ->
```
vasilisp: @Drup, this should be `"shared.start"|"client.start"|"server.start"` (consistently with structures and the documentation), right?

vasilisp:
As a further test, I converted `ocsigen-toolkit` to PPX: ocsigen/ocsigen-toolkit@53b141961eb1008a74a550b4a63665e814aa5e27 . The conversion was trouble-free, modulo small glitches with the `.eliomi` files (which are easy to fix).

src/ppx/ppx_eliom_utils.ml
Drup: Indeed.

Drup: Fixed.

Drup:
@vasilisp Great. I think we are pretty much ready to merge.

vasilisp:
@Drup, can we merge? If any issues arise, we can fix them in master.

My mind cannot handle too much pinning and too many switches :).

Drup:
Absolutely. I was expecting you to do when done with testing. :)

PR #215, opened 2015-12-02
Improved Form APIs

vasilisp:
This PR re-structures our form APIs. This is a breaking change.

The form APIs are now available under `Eliom_content.Html5.{D,F}.Form`. `a` and related functions have not moved.

The functions `${t}_${formtype}` (e.g., `int_input`) have been replaced by `${formtype}` functions with an additional "GADT" argument denoting the type. Internally, this is implemented with `Eliom_parameter_base.atom`. The conversion is almost mechanical; see ocsigen/eliom-base-app@ae247f1d2a210bef5d4b85cb2e6b5c62596f62a6.

The `user_type_*` functions have gone away. I don't think they are used all that much, and we probably need to re-implement them in a safe fashion anyway.

`Eliom_form` has been simplified internally and merged with `Eliom_mkforms`.


Drup:
I like the new API, but I'm not fond of breaking every eliom code in existence. 

How much more complexity would it add to keep the old one, with a deprecation notice ?

vasilisp:
Roughly speaking, we would need to re-add the functions removed by 41d3969 and 4f7688d as a separate functor, and instantiate it to produce the old API.

But to be honest, I think we should feel free to break backwards compatibility when this simplifies the APIs and the internals :). The userbase is not that huge, and temporarily maintaining the deprecated APIs adds overhead to the (needed) Eliom cleanup/simplification process.

Drup:
Do we have anything preventing merging ? The migration page need to be done, but it's independent.

vasilisp:
It is mostly about testing and updating a few notable projects that use Eliom forms, I guess. It doesn't get in the way of anything else we are doing.

Drup:
@vouillon @balat : Did you tested this ?

vasilisp:
I realized that I hadn't run the test suite on this branch. Updating for the new API is not 100% smooth, but working on it. (This is a blocking issue.)

balat:
Hello,
I'm currently testing this, and it not difficult to do the switch, provided there are migration instructions (see bottom of page http://ocsigen.org/eliom/install ).
Also I prefer to be sure that all the documentation has been updated before merging this:
- [ ] Eliom manual
- [ ] API
- [ ] Tutorial
- [ ] Mini HOWTOs

What is the status of all this?
Vincent

vasilisp:
The API documentation is inline and it should be up-to-date. The manual is in progress. I will deal with the tutorial and the mini HOWTOs once the manual is done.

vasilisp:
I have updated all the documentation I am aware of (by fairly blindly transforming the old API to the new). I will do another pass once the `gadt-forms` branch gets merged. I have hard time getting the markup right and I need to see how it looks :).

I will provide migration instructions.

balat:
Great!
One more question: ```button ~button_type ...``` became ```button ~a:[a_button_type ...]``` (but not for inputs). Does it mean that the type attribute is not mandatory any more for ```<button>``` ?

vasilisp:
@balat, good catch. It is a mistake, I didn't eliminate the `~button_type` argument on purpose. I will add it back.

balat:
And what do we do with inputs?

vasilisp:
There is an `~input_type` argument like before. It doesn't work?

balat:
Yes it works sorry.
I just re-translated bs without major issue.
Actually, the only small issue I had is that there exists a module called Form in js_of_ocaml ...
I'm merging the PR.

darioteixeira:
Have you considered the use case where the input type is abstract? Consider for example the approach used by [`Litiom_type`](https://github.com/darioteixeira/litiom/blob/master/src/litiom_type.mli).

vasilisp:
@darioteixeira, `Litiom_type` gives us ideas for improvements, thanks!

You can use the new interface with an abstract input type `t` such that you have a `t -> t param` function, so I think we have made a step forward (but it is not perfect yet).

With some effort we can probably turn `param_name` into a GADT which can take over the role that `'a param` has in `Eliom_form`. We will then be able to get rid of the `'a param` argument that I added to `input`, `button` & co. (like in `Litiom_type` but without an additional layer).

darioteixeira:
@vasilisp: The main goal of Litiom is to make itself obsolete by providing ideas to Eliom. Therefore, if Eliom 5 offers convenient support for better abstraction in service parameters and forms, I'll be happy... :-)

Anyway, I haven't had the chance to play with the dev version of Eliom yet.  When I do I'll be able to provide better feedback.

darioteixeira:
I'm trying to port Litiom to Eliom5, and I've noticed that `user_type_input` and related functions have disappeared.  I may be missing something obvious, but how does one produce a form for a user-defined type in Eliom5?

PR #218, opened 2015-12-10
Update Eliommod_timeouts interface

vasilisp:
This patch re-organizes `Eliommod_timeouts`.

Previously, there were separate `*_data_*`, `*_service_*`, and `*_persistent_*` functions. These have been unified by means of a parameter of type ``[ `Service | `Data | `Persistent ]``. I dropped the redundant `_timeout` suffix from everything.

I intend to do similar re-structuring in `Eliom_state` which wraps this interface, but let's review things incrementally.

### DIFF
--- a/src/lib/eliom_state.server.ml
+++ b/src/lib/eliom_state.server.ml
@@ -82,7 +82,7 @@ let set_default_global_service_state_timeout ~cookie_level
   let sitedata =
     Eliom_request_info.find_sitedata "set_global_service_timeout"
   in
-  Eliommod_timeouts.set_default_global_service_timeout
+  Eliommod_timeouts.set_default_global `Service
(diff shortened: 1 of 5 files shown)


Drup:
Should we also have `` `Volatile`` and only accept it in some functions ?

vasilisp:
I am more inclined to just remove the single `*volatile*` function. It is a trivial sequencing of two others, and it can happen at a higher-level module.

Drup:
@vasilisp Do we merge this now or later ?

vasilisp:
While it is more meaningful in combination with what happens in the `state` branch, even in isolation the `Eliommod_timeouts` cleanup doesn't harm. So let's get it out of the way.

Drup:
Agreed.

PR #233, opened 2016-01-19
Many changes aiming at reducing the size of request data

vouillon:


Drup:
I was looking at the client/server syntax "runtime" last week, and I was making myself familiar with it before doing a big cleanup. This is very welcome, I don't have to do the clean up anymore! :D

Give me tomorrow to look at it in more details, so that I can convince myself we are not breaking anything (especially the revised implementation of client values). 

Please add some comments for implementation decisions.

Drup:
@vouillon Did you measured the improvements ? Can you give us an idea ?

vouillon:
For the Besport landing page, the uncompressed size of the request data goes from 130k to 78k (40% lower), and the compressed size from 24755 bytes down to 17059 bytes (30% lower).

vouillon:
I added some comments.

Drup:
@vouillon You are using bytes apparently, which doesn't exist in old ocaml. We will need to use base-bytes and modify this a bit.

PR #235, opened 2016-01-20
Documentation improvements

vouillon:
- [ ] Still needs to describe the compilation process with `ppx`.
- [ ] Maybe `basic.ppx` should be the default template for `eliom-distillery`?

### DIFF
--- a/doc/manual-wiki/examples.wiki
+++ b/doc/manual-wiki/examples.wiki
@@ -1,5 +1,0 @@
-=@@id="examples"@@Demos=
-
-Go to the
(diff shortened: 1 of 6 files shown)


vasilisp:
We have used PPX enough to know that it works,  and it is better than the Camlp4 syntax anyway, so I am in favor of switching by default. Some issues will arise, e.g., somebody will try to put PGOCaml or Macaque Camlp4 syntax in Eliom files, but delaying the switch is not the solution.

vasilisp:
This looks good, and the PPX points are quite orthogonal, so merging.

PR #236, opened 2016-01-21
Tools to inspect request data

vouillon:


Drup:
I guess you didn't wanted to fiddle with the build system, but we should build a single executable instead of two scripts. It's not very important, though.

PR #237, opened 2016-01-21
Require bytes for the syntax extension.

Drup:
@hhugo: Can you check ? I think it's correct, but I had to struggle with ocamlfind a bit ...


Drup:
Should be better now.

vasilisp:
I don't understand the `P s -> Quote (A s)` transformation` but you must have a reason :). Otherwise it looks OK.

Works for me on 4.01 and 4.02. Merge?

Drup:
@vasilisp I don't either, but ocamlbuild seems to try to do magic with `P` elements, and I want to avoid that.

PR #238, opened 2016-01-22
Fix ppx for 4.03

Drup:
Only the ppx was fixed. In particular, pa_include doesn't work.


vasilisp:
The code looks good to me. My 4.02 switch doesn't care at all about these changes.

Drup:
I'm merging, but we are not ready for 4.03 yet, unfortunately, because of the camlp4 part (and I think we will need some preprocessing to handle it)

PR #239, opened 2016-01-22
Use the ppx template by default

Drup:
Here we go. :)


vasilisp:
Shouldn't `basic` refer to what is now `basic.ppx`? With a `basic.camlp4` and a possible alias between `basic` ~ `basic.ppx`.

(Details, I am happy enough if PPX is the default and what we promote, no matter what the names are.)

Drup:
I'm ok with renaming `basic` to `basic.camlp4`. I don't feel the need for an alias.

vasilisp:
Good, with `basic.ppx` and `basic.camlp4` there is no room for confusion about what syntax is supported.

Drup:
@vasilisp Done.

PR #263, opened 2016-02-16
Restore user parameters in forms

### LINKED ISSUE #256: Custom types for forms
Can we please get back a functional equivalent of `user_type_select` and friends in Eliom 5.0+?  The param type used in the forms is aliased to `Eliom_parameter_base.atom`, but in that module only the `params_type` admits user defined types.  Would it be acceptable to add a `TUserAtom` constructor for `params` holding converter pair and adding something like

```ocaml
val user : to_string: ('a -> string) -> of_string: (string -> 'a) -> 'a param
```

to the Form signature?  I'm naming things here just so I can send a pull request if you agree.

vasilisp:
A re-implementation of user-type form parameters, which had gone away with #215.

I hope this is temporary, with a better solution (integrated with `Eliom_parameter`) coming in the future.

Fixes #256. @darioteixeira @paurkedal, sorry for the inconvenience.

### DIFF
--- a/src/lib/eliom_content_.client.ml
+++ b/src/lib/eliom_content_.client.ml
@@ -425,7 +425,7 @@ module Html5 = struct
   type +'a elt = 'a F.elt
   type +'a attrib = 'a F.attrib
   type uri = F.uri
-  type 'a form_param = 'a Eliom_parameter_base.atom
+  type 'a form_param = 'a Eliom_form.param
 
   module Custom_data = Eliom_content_core.Html5.Custom_data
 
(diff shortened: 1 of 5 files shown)


darioteixeira:
Thanks @vasilisp!

paurkedal:
Thanks, that should help a lot for now.  I only inspected it though, as I don't seem have the right build environment for this branch.  I taken another look if you need testing.

vasilisp:
@paurkedal, thanks, I did some testing myself, and don't think much testing is needed. For the record, you currently need TyXML from master to compile Eliom from master; same for this branch.

@Drup, any comment?

Drup:
@vasilisp Not really, it looks fine to me, but I don't know this part all that well.

vasilisp:
Good, merging.

paurkedal:
@vasilisp Yes, looked like TyXML, and wikidoc not in opam yet, and then it was getting late. Thanks for the user parameters!

PR #268, opened 2016-02-17
PPX support in Ocamlbuild

### LINKED ISSUE #257: ppx support in ocamlbuild_eliom
`ocamlbuild_eliom.ml` currently hardcodes the camlp4o preprocessor option and related package names. I think it could support ppx e.g. via a global ref which could be set `Before_options` without to intrusive changes.

vasilisp:
PPX support for our Ocamlbuild plugin.

You can use PPX by using the `eliom_ppx` in flags as follows:

```
<*.eliom>: eliom_ppx
<*.eliomi>: eliom_ppx
```

Fixes #257. The plugin has an early adopter already; thanks @paurkedal!


Drup:
Looks good to me !

Issue #277, opened 2016-03-02
eliom can't be built with camlp4.4.02+7


kandu:
after the upgrade of camlp4 from 4.02+6 to 4.02+7

```
kandu@bomb:~$ opam install eliom
The following actions will be performed:
  ∗  install eliom 5.0.0

=-=- Gathering sources =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[eliom] Archive in cache

=-=- Processing actions -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[ERROR] The compilation of eliom failed at "make".

#=== ERROR while installing eliom.5.0.0 =======================================#
# opam-version 1.2.2
# os           linux
# command      make
# path         /home/kandu/.opam/4.02.3/build/eliom.5.0.0
# compiler     4.02.3
# exit-code    2
# env-file     /home/kandu/.opam/4.02.3/build/eliom.5.0.0/eliom-23351-1872ec.env
# stdout-file  /home/kandu/.opam/4.02.3/build/eliom.5.0.0/eliom-23351-1872ec.out
# stderr-file  /home/kandu/.opam/4.02.3/build/eliom.5.0.0/eliom-23351-1872ec.err
### stdout ###
# [...]
# File "src/syntax/pa_eliom_seed.ml", line 815, characters 22-23:
# While expanding quotation "expr" in a position of "expr":
#   Parse error: [record_field] or [expr level .] expected after "{" (in [expr])
# 
# File "src/syntax/pa_eliom_seed.ml", line 1:
# Error: Error while running external preprocessor
# Command line: camlp4 '-I' '/home/kandu/.opam/4.02.3/lib/ocaml/camlp4' '-I' '/home/kandu/.opam/4.02.3/lib/ocaml/camlp4' '-I' '/home/kandu/.opam/4.02.3/lib/ocaml/camlp4' '-parser' 'o' '-parser' 'op' '-printer' 'p' '-parser' 'Camlp4QuotationCommon' '-parser' 'Camlp4OCamlOriginalQuotationExpander' '-parser' 'Camlp4GrammarParser'  'src/syntax/pa_eliom_seed.ml' > /tmp/ocamlpp06b533
# 
# Command exited with code 2.
# Makefile:13: recipe for target 'all' failed
### stderr ###
[...]

PR #278, opened 2016-03-02
Compile against Camlp4 4.02+7

vasilisp:
The just-released Camlp4 4.02+7 modifies the grammar entries for records. (See the series of commits that fix ocaml/camlp4#102.) Given that Eliom adjusts these to not conflict with our uses of the brackets, we have to catch up.

The rule deletion part is far from pretty. We delete a different rule depending on the version of Camlp4 installed. I can't claim I understand how things work, but I tested with both OCaml 4.01.0 (with assorted Camlp4) and OCaml 4.02.3 with new Camlp4, and everything seems to be fine.


vasilisp:
Fixes #277 .

PR #279, opened 2016-03-03
Removing Eliom_pervasives and make it possible to use syntax in Eliom_service

balat:


vasilisp:
There is `Eliom_client_common_base`, `Eliom_client_common`, `Eliom_client_common2`, `Eliom_client0`, and `Eliom_client`. What is the purpose of each of these?

If there is some logical separation, it should be reflected in the names. Otherwise, we are bound to make the codebase even more unmaintainable, because it will never be clear where to put what.

vasilisp:
The sooner we merge this, the better. But let's fix the naming scheme first. What about the following:

```
Eliom_client_common2     -> Eliom_client_syntax
Eliom_client_common      -> Eliom_client_value
Eliom_client_common_base -> Eliom_client_value_base
Eliom_client0            -> Eliom_client_core
```

`Eliom_client` looks OK to me.

The `_core` suffix doesn't suggest sharing, and it is in alignment with `Eliom_client_core`.

What is currently `Eliom_client_common_base.RawXML` will ideally go to the new `Eliom_client_core`. The "shared" part of `Eliom_client_value` will move to `Eliom_shared`. These cleanup steps can happen later.

vasilisp:
b2d1699 is not needed. It was part of a custom solution for hot code pushing, but we will be using a Cordova plugin instead. I am planning to rebase it out.

vasilisp:
Please express any concerns on the names today :). The renaming effort spans across a few repositories and it would be a waste of time to change it multiple times.

Drup:
At this point, this is pretty much impossible to review (too many loosely related changes with big diffs), so I'm just going to trust you. 

@vasilisp Why client_value and client_value_base ?

vasilisp:
@Drup, it is a user-visible module containing many "Eliom runtime" things, but the users only (?) need to refer to what is now `Eliom_client_common.client_value` (currently the "canonical" name of `client_value`). That would become `Eliom_client_value.t`, which I find agreeable.

`Eliom_runtime` would be technically more accurate, but a name like `Eliom_runtime.client_value` looks very "leaky" to me. Providing an `Eliom_runtime.client_value` alias elsewhere is to be avoided, because our current type checking doesn't like synonyms of `client_value`.

It is unfortunate that there are other things there not directly related to client values. We can hide them from the docs for now, and incrementally end up with a better organization. Anyway, I know it is not a perfect solution, and I am open to suggestions :).

Drup:
I agree with having `Eliom_client_value` (I have exactly that in my version of the runtime). I was just wondering about the `base` module. 

I have an `Eliom_runtime` in my version, for all the syntax extension things.

vasilisp:
I just reproduced the scheme that existed before, but you have a point. We can de-couple the `base`, i.e., `Eliom_client_common_base -> Eliom_runtime` (referred-to internally) and `Eliom_client_common -> Eliom_client_value` (without including `Eliom_runtime`).

Drup:
@vasilisp Yes, that sounds good.

vasilisp:
The renaming has happened in the `client-module-names` branch. I am now going to update all the reverse-dependencies of Eliom to make sure everything is fine.

vasilisp:
I pushed `ocsigen-widgets`, `ocsigen-toolkit`, and `eliom-base-app` branches that use the new names. All branches are called `client-module-names`.

@balat @vouillon, as soon as you migrate to these, we can merge everything :).

balat:
@vasilisp We are now using the new names. I'm ok to merge everything.

vasilisp:
I am glad :). I will take care of the merging.

vasilisp:
Done, you can start using `master` for everything (except TyXML).

PR #289, opened 2016-04-14
Fix Eliom_service.ocaml_service variance

vouillon:
The type cannot be covariant as this would make the type of function `Eliom_registration.Ocaml.register` unsound.


Drup:
Hum, can you explain ?

PR #291, opened 2016-04-14
Remove Eliom_parameter.TCoordv

vasilisp:
This PR removes the `Eliom_parameter.TCoordv` constructor, which is for coordinates that carry an additional value. Such coordinates could be produced by form elements like the following:

```html
<input type="image" value="foo" />
```

However, the above is not of much use, it is [not universally supported](https://stackoverflow.com/questions/7935456/input-type-image-submit-form-value) by browsers, and it is no longer supported by `Eliom_form`. So let's remove some cruft.

### DIFF
--- a/src/lib/eliom_parameter.server.ml
+++ b/src/lib/eliom_parameter.server.ml
@@ -161,13 +161,6 @@ let reconstruct_params_
          | r, l ->
            let rr, ll = parse_suffix (TAtom ("",TInt)) l in
            {abscissa = r; ordinate=rr}, ll)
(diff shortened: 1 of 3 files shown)


vasilisp:
Actually, I would like to remove some other kinds of parameters. Off the top of my head:

* `TList`: we have sets
* `TTypeFilter`: does the filtering really need to happen inside `Eliom_parameter`?
* `TOption (_, b)`: do we really need multiple semantics for options (`b` value, `neopt` and `opt` helpers)?

PR #292, opened 2016-04-14
Eliom loader: wait for updates; add retry button on update failure

vouillon:


vasilisp:
Merged. I think it would be nice to make the Hot-Code-Push plugin optional; the rest of the script doesn't even assume Cordova. But we can do this later.

Issue #209, opened 2015-11-06
Simplifying Eliom service interface


balat:
There is a simplification that I have been wanting to do for years in the interface of Eliom_service and that is mostly possible now (I think) thanks to GADTs.

The idea would be mainly to have only one function to create a service, with some parameters to decide the kind of identification we want (path, parameter or both) and the method (GET/POST).

The main difficulty is that these different kinds of services do not take exactly the same parameters.
For example POST services take a GET service as fallback, etc.

Any idea welcome :)

vasilisp:
A solution would be to make the fallback part of the "identification" data structure (for the applicable cases). The types become scary because the fallback belongs in the type `service` (which has 10 type parameters), but we will simplify things (*somehow*).

vasilisp:
It took quite a while, but the `service` branch now has an all-encompassing [`service` function](https://github.com/ocsigen/eliom/blob/544427efb6c707ea9fc7fbf4481c616219ce0ed9/src/lib/eliom_service_sigs.shared.mli#L104).

Drup:
```ocaml
    meth:('m, 'gp, 'gn, 'pp, 'pn, 'tipo, 'mf, 'gp_) meth ->
    id:('att, 'co, 'mf, 'rt, 'gp_) id ->
    unit ->
    ('gp, 'pp, 'm, 'att, 'co, non_ext, reg, 'tipo, 'gn, 'pn, 'rt) service
```

Blurb, that's not very digestible. ;)

vasilisp:
`meth` should be comprehensible: the method is one of `Get`, `Post`, `Put`, and `Delete`, and all these come with `GET` parameters, except `POST` which also comes with `POST` parameters. 

`id` isn't complicated either: a service is either located on a path `p` (`Path p`), or it sits on top of a `fallback` service (`Overlay fallback`), or it answers on all URLs thanks to service identification magic (`Global`).

So basically there is almost no redundancy in the information you need to provide.

The type parameters were there before, and they can't really go away :).

Drup:
Oh, I agree that the *constructors* are fine, it's the *types* that are quite scary. Unfortunately, OCaml users are used to be directed by types. But I'm sure documentation will be enough. :)

vasilisp:
I understand the objection, and I am not even sure that documentation is a sufficient remedy.
However, there are good reasons why all these parameters are there. Let me write down my view of them in case this triggers any ideas :).

[...]

Issue #223, opened 2015-12-18
Incorrect type definition of services


vouillon:
In the type declaration:
```
type ('get,'post,+'meth,+'attached,+'kind,+'tipo,'getnames,'postnames,+'registr,+'return) service
```
the variance of the last parameter is wrong.
Indeed, if the service is defined in two step as below, the parameter remains polymorphic in the type of the service `service` instead of being refined by the type of `f`.
```
let service =
  Eliom_service.Ocaml.post_coservice'
    ~name:"service"
    ~post_params:(Eliom_parameter.file "f")
    ()
let () = Eliom_registration.Ocaml.register ~service f
```
We need to check that we do not break anything by removing the variance annotation, though.

vasilisp:
I pushed `service_variance` branches for Eliom and EBA, and their combination builds.

vasilisp:
I think this will be fixed in the process of cleaning up the service APIs. The complicated type parameters get in the way of providing a more uniform interface (#209).

Additionally, the last parameter is what necessitates `sigs/eliom_service.mli`, which we need to get rid of in preparation for a PPX future (#200) :).

vasilisp:
In a related note, does anyone know/remember why the last parameter of a fallback service can differ from the last parameter of the service produced (e.g., [in `Eliom_service.Http`](https://github.com/ocsigen/eliom/blob/master/src/lib/eliom_service.server.mli#L177))? Eliminating the distinction will help.

vasilisp:
The types have changed, and there aren't variance annotations any more :).

PR #293, opened 2016-04-15
GADT-ify Eliom_service API

vasilisp:
This PR unifies the `Eliom_service` (and `Eliom_registration`) API by using GADTs. We now have one main entry point, `Eliom_service.service`, which can be used as follows:

```ocaml
Eliom_service.service ~meth ~id ~rt ()
```

`meth` is one of:

* `Eliom_service.Get get_params`
* `Eliom_service.Post (get_params, post_params)`
* `Eliom_service.Put get_params`
* `Eliom_service.Delete get_params`

`id` is one of:

* `Eliom_service.Path path` (for what used to be a regular service)
* `Eliom_service.Fallback fallback` (for what used to be an attached coservice)
* `Eliom_service.Global` (for what used to be a non-attached coservice (`coservice'`)

`rt` is one of `OCaml`, `Http`, `Appl`, and `Unsafe`. The latter accommodates some very weird services allowed by `Eliom_registration`. I hope it will go away at some point.

A fallback for a POST service can no longer get `GET` parameters. This was a necessary trade-off to allow for the unification of the API. Also, the `PUT` and `DELETE` services no longer get `raw_post_data`.

A significant amount of boilerplate has been removed. As a side-effect, we are almost ready to get rid of `pa_include`. It is now only used to share `eliom_reg_simpl.mli`, which has become very short (3 functions), so we could just copy it where needed (in the few cases in `Eliom_registration` where standard OCaml signature inclusion doesn't work).

This requires fairly rigorous review, including by myself.


vasilisp:
The PR will close #209, and there is some related discussion there.

Drup:
:yay: 

I took a look at eliom_parameter(_base). There is still a lot of duplication and unused functions. I shaved off 200 lines, but we can remove a lot more by just removing the so many functions that are exposed but undocumented ...

There are some unused `priority` arguments here and there, is that intended ?
We really need to enable more warnings.

Drup:
There is `pre_wrap` function in `eliom_service.client` and `eliom_service_base` (and they are not identical ...).

vasilisp:
> I took a look at eliom_parameter(_base). There is still a lot of duplication and unused functions. I shaved off 200 lines, but we can remove a lot more by just removing the so many functions that are exposed but undocumented ...

Good job with the boilerplate, thanks. I ran out of steam and didn't go all the way :). 

You mean `Eliom_service` (not `parameter`), right? I do hope that a cleanup pass over `Eliom_parameter` will follow.

> There are some unused priority arguments here and there, is that intended ?

No, I will fix it.

> We really need to enable more warnings.

I for one do need to get in the habit of developing with more warnings :).

vasilisp:
> There is `pre_wrap` function in `eliom_service.client` and `eliom_service_base` (and they are not identical ...).

The one in `eliom_service.server.ml` was not used internally, but it was exported (no clue why, it could even be `ocamlc -i`). The one in `Eliom_service_base` looks fine to me, so keeping that for internal use.

vasilisp:
I am considering switching to the `t` convention everywhere, i.e., `Eliom_service.t` (for what is now the type `service`), `Eliom_service.Id.t`, `Eliom_service.Ret.t`. Also, I would call the main functions `Eliom_service.make` and `Eliom_registration.X.make` (we will still have `Eliom_registration.X.register`). Opinions?

I would like to merge the branch ASAP. It is hard to maintain as a branch, we can improve the API later, and there are other things that need to be done in Eliom.

Drup:
> I am considering switching to the t convention everywhere

Yes please. I would rather use `create` than `make`, though.

vasilisp:
Done. `create` it is.

`ocsigen-toolkit`, `ocsigen-widgets` and `eliom-base-app` are all up-to-date (`service` branches).

src/lib/eliom_comet.server.ml
```diff
@@ -21,9 +21,8 @@
 
 (* TODO: handle ended stream ( and on client side too ) *)
 
-open Eliom_lib
-
 module Ecb = Eliom_comet_base
+let (>>=) = Lwt.(>>=)
```
Drup: I encourage using `open Lwt.Infix` instead.

src/lib/eliom_service_sigs.shared.mli
```diff
@@ -129,60 +133,57 @@ module type S = sig
...
 
       The whole URL is constructed from the prefix, the path and GET
       parameters. Hence, an empty prefix can be used to make a link to
       another site of the same server.
 
       See {!val:service} for a description of the optional
-      [~keep_nl_params] and [~rt] parameters.  *)
-  val external_service :
+      [~keep_nl_params] and [~ret] parameters.  *)
+  val create_external :
```
Drup: Maybe just `external` ?

src/lib/eliom_service_sigs.shared.mli
```diff
@@ -283,27 +277,26 @@ module type S = sig
...
 
   val get_info :
-    (_, _, _, 'att, _, _, _, _, _, _, _) service ->
-    'att attached_info
+    (_, _, _, 'att, _, _, _, _, _, _, _) t -> 'att attached_info
 
-  val is_external : (_, _, _, _, _, _, _, _, _, _, _) service -> bool
+  val is_external : (_, _, _, _, _, _, _, _, _, _, _) t -> bool
 
   val get_get_params_type_ :
```
Drup: We could drop the `get_` for all those.

src/lib/eliom_service_sigs.shared.mli
hhugo: `external` is a keyword.

Drup: Oh. :(

vasilisp: The implicit association with `create` doesn't harm :). I even thought about merging `create_external` into `create` (with another `Id.t` constructor), but I decided against complicating the `Id` type a bit more for a case of limited use.

Drup: Does it really makes things more complicated ? Is it not just another case in Id ?

src/lib/eliom_service_sigs.shared.mli
vasilisp: And the `_` suffix. Haven't bothered yet because these are exclusively for internal use.

src/lib/eliom_service_sigs.shared.mli
vasilisp: Just another case operationally speaking, but with additional type parameters. It would add an `ext` type parameter to `Id.t`, and I think a `reg` type also.

Drup: > but with additional type parameters

grmph, that is very unfortunate indeed.

balat:
@darioteixeira Do you have any coments about this?

vasilisp:
I am in the process of updating the testsuite. There are quite a few services there :). The testsuite gives raise to the following point.

I am thinking of providing an `Eliom_registration.X.create_post_with_fallback` (not sure about the name) that restores the behavior of the old `Eliom_registration.X.register_post_service` (not coservice) that used to take a fallback. This will be entirely in `Eliom_registration`, implemented in terms of the GADT `Eliom_service` API.

_Motivation_: it is useful to associate a POST service to a GET service on the same URL, e.g., for bookmarking. However, this is not a strict requirement (POST services without fallback are possible), and it breaks the symmetry that enables the new API, hence it can be delegated to a separate function.

balat:
I started translating BeSport to the new interface.
 - I had a problem with an Http coservice that was attached to an Appl service. It is a service that goes to the main page of the site, and disconnect the user. This was a way to force restarting the app. Is it still possible to do that or do you need the same type variable 'ret for both services? I solved the problem by using Eliom_client.exit_to instead of Eliom_client.change_page. But that's more risky.
 - More problematic: Actions always have return type Http, if I'm right (or is there a way to make an action inside the app?). This is an issue because all action will restart the app ... (?) 
 - I suggest to rename `Eliom_service.Ret.Appl` to `Eliom_service.Ret.App`, for uniformity (and because we says "an app".

(I'll continue tomorrow)

vasilisp:
> I had a problem with an Http coservice that was attached to an Appl service. It is a service that goes to the main page of the site, and disconnect the user. This was a way to force restarting the app. Is it still possible to do that or do you need the same type variable 'ret for both services? I solved the problem by using Eliom_client.exit_to instead of Eliom_client.change_page. But that's more risky.

I [was wondering](https://github.com/ocsigen/eliom/issues/223#issuecomment-179190197) whether the difference in return types is useful :). I think we can restore this with another type parameter in `Ret.t`.

> More problematic: Actions always have return type Http, if I'm right (or is there a way to make an action inside the app?). This is an issue because all action will restart the app ... (?) 

If we turn it into `App`, does that cover all needs?

> I suggest to rename Eliom_service.Ret.Appl to Eliom_service.Ret.App, for uniformity (and because we says "an app".

Sure.

vasilisp:
@balat, I do not understand actions very well. Does it suffice if the fallback service of the action is an "app" service?

vasilisp:
I have a hard time understanding the old `Eliom_registration` API. In particular, why do we have a `` `returnT `` in [https://github.com/ocsigen/eliom/blob/master/src/lib/sigs/eliom_reg_simpl.mli#L49](`register`), which is the same as for fallbacks, instead of a `` `returnB ``?

darioteixeira:
Thanks @balat for CCing me on this.  I'll make some experiments and report back on the API.

At first glance the only thing that comes to mind is that `Global` doesn't seem very descriptive of non-attached co-services.  Sure, they are global in a sense, but if you asked me to describe *non-attached co-services* in a single word I would probably say `Unattached`...

balat:
I'm also trying to understand. Actually, what choses if the service stays in the app or not is the send_appl_content field, which seems not to be related to the `return` type parameter ...
It was already the case that services created using `Eliom_registration.Action` had return type http ...
But indeed, it was possible to register an action on a service declared to be "app" or "http".
I would say that current situation is ok ...
But we need to understand better the need for this return type parameter.
What if we merge App and Http?
It is really used as a constraint in `Eliom_client.call_ocaml_service`, because we need to know the type of the ocaml data returned.

vasilisp:
> What if we merge App and Http?

> It is really used as a constraint in Eliom_client.call_ocaml_service, because we need to know the type of the ocaml data returned.

That's my understanding as well. Given the type mismatch between `register` and `register_service` (et al.), there was not clear distinction between `App` and `Http`. It is the difference with Ocaml services that is needed (and straightforward). But yes, we need to be careful.

balat:
I just finished to switch BS. It seems to be ok now :)

vasilisp:
Awesome. I think the remaining issues can be handled with other PRs. Any objections to merging?

balat:
No objection :)

darioteixeira:
The API documentation hasn't been fully updated yet, which makes exploration a bit harder.  Nevertheless, I managed to get a simple hello world to type-check and compile:

````ocaml
let hello_handler () () =
    Lwt.return
        (html
        (head (title (pcdata "Hello")) [])
        (body [p [pcdata "Hello World!"]]))

let hello_service =
    Eliom_service.(create
        ~ret:Ret.Non_ocaml
        ~meth:Meth.(Get Eliom_parameter.unit)
        ~id:Id.(Path [])
        ())

let () =
    Eliom_registration.Html5.register hello_service hello_handler
````
However, it fails at runtime with the message "This function cannot be called here because it needs information about the request or the site".  Any idea of what's wrong?

vasilisp:
@darioteixeira, sorry for the late reply. I came upon the same issue while trying to restore the testsuite. I think it is fixed in master (https://github.com/ocsigen/eliom/commit/38f2f1ace81de0ae189bb27c94610a38b78a32b3).

darioteixeira:
@vasilisp: Yeap, it works now, so that was probably it.  I can now play with Eliom 6.0... \o/

Issue #294, opened 2016-04-21
Discussion on Eliom6 naming


darioteixeira:
I've started playing a bit with the upcoming Eliom6.  Since bikeshedding season is officially open, here are some initial thoughts on the API (more to come later):

* The variants in `Eliom_service.Ret.t` are `Ocaml | Non_ocaml | Unsafe`.  I think it's generally preferable to name a variant by what it **is** instead of that it's **not**.  Doesn't `Non_ocaml` pertain to services which return non-executable data such as Html, raw text, etc?  Wouldn't it be better then for the variants to be named `Code | Data | Unsafe`?

* The constructors are named `create`, which goes against the community trend of calling them `make`.  Even the Array module in the Stdlib has deprecated `create` in favour of `make`...

* The new API depends heavily on constructor disambiguation if you want to avoid lots of parentheses and full qualifications. I understand there may be strong reasons to declare `Id`, `Ret`, and `Meth` as submodules, but without relying on disambiguation this approach leads to a very messy declaration of even a simple service:

````ocaml
let hello_service =
    Eliom_service.(create
        ~ret:Ret.Non_ocaml
        ~meth:Meth.(Get Eliom_parameter.unit)
        ~id:Id.(Path [])
        ())
````

Drup:
> Code | Data | Unsafe?

I don't disagree, but that's wrong, all of them are data. Just (almost raw) OCaml objects or more standard objects. It's more a question of how things are serialized actually, either text or marshal, but I really don't want to put `Marshall` in the API, it's really an implementation detail.

> against the community trend of calling them make.

There is no community trend, it's all over the place. To counter your argument, `Hashtbl.create`, `Bigarray.X.create`, `String.create` (which is not the same as `String.make`, etc.

> The new API depends heavily on constructor disambiguation

Huum, that's a good point, maybe we could reexport them inside `Eliom_service` ?

vasilisp:
The submodules have very little code, and that code is only needed inside Eliom. They are only there to prettify the namespace, but apparently there are counter-arguments :). I would rather remove them than re-export them; I much prefer a single canonical name to refer to things.

Drup:
> I would rather remove them than re-export them; I much prefer a single canonical name to refer to things.

I disagree a bit. `type meth = Meth.t = ...` is a common pattern.

darioteixeira:
[...]

PR #295, opened 2016-04-22
Eliom_service and Eliom_registration adjustments

### LINKED ISSUE #200: Deprecate pa_include and remove the sigs directory

### LINKED ISSUE #285: Service registered by Eliom_registration.App breaks co-located apps
`Eliom_registration.App` [registers a service](https://github.com/ocsigen/eliom/blob/master/src/lib/eliom_registration.server.ml#L2210) for accessing the application's global data. This is to enable global injections in mobile apps. Currently, this is a GET service in a hard-coded location. If the Ocsigenserver runs with multiple Eliom apps (e.g., as is the case for the testsuite), we try to register multiple services on the same URL (and clearly fail).

We will have to make the URL configurable, or at least make the service optional.Unfortunately, this will be a breaking change requiring a one-line modification in all existing Eliom applications.

vasilisp:
The PR contains a few adjustments related to services and registration. They are grouped to ease the transition for the users of `master`.

### GADT service API

The `Meth` and `Id` submodules have gone away to make the constructors available directly under `Eliom_service` (see #294). We could keep the submodules and re-export, but I see no concrete use for them.

The `~ret` argument no longer exists. `Eliom_service.create` handles the standard case, the other cases (mostly for internal use) being available as `Eliom_service.create_{unsafe,ocaml}`. This eases our work with the documentation. The code duplication needed is basically only in the signature.

External services are now handled via a `meth` constructor.

The `Eliom_service_types` module is replaced by `Eliom_service_sigs.TYPES`.

### Global data service

It is now optional. Fixes #285.

### `Eliom_registration` signatures reorganized

Fixes #200.


Drup:
I still think we should keep the submodules. The API is nicer with namespaces like that, and it means we can add new convenience functions.

vasilisp:
I disagree :).

The constructors are the only part of the API that is interesting to users, and it is better (as per #294) if they are available under `Eliom_service`. Why provide redirection if it is the `Eliom_service` names are the ones that we really want?

There is very little that we can add. These types have a very specific role (specifying the service parameters), and they are basically feature-complete.

Drup:
```ocaml
type id = ...

(* At the bottom of the file *)

module Id : sig
  type t = id = ...
  val untype : ...
end
```

The user doesn't have to look inside it and the various functions are not spilled inside `Eliom_registration`, but it's still properly grouped and documented.

vasilisp:
So, a module to group `id` with a single (and trivial) function that doesn't even need to be documented, because it only exists to avoid an `Obj.magic` [in `Eliom_registration`](https://github.com/ocsigen/eliom/blob/service/src/lib/eliom_registration.server.ml#L984)? Come on, let's not exaggerate :).

Drup:
I think we're not going to agree, so I'll let you do it and rant when we need to add new functions that are also going to end in `_id`. :p

PR #297, opened 2016-05-05
Make it compilable with Ocaml 4.03.0.

tobiasBora:
Ocaml 4.03.0 seems to be a little bit stricter, these minor changes are usefull to make it compile on Ocaml 4.03.0.


vasilisp:
Thanks! I am surprised that you got that far. Are all the dependencies ready for 4.03 (e.g., Ocamlnet)?

Looks fine to me (we export the exception via the preceding signature inclusion), so merging.

tobiasBora:
To compiled Ocamlnet I used the branch "fixes-ocaml-4.03" here : "https://gitlab.camlcity.org/gerd/lib-ocamlnet3". However now I have problems with ocsigen-toolkit:
https://github.com/ocsigen/ocsigen-toolkit/issues/17

PR #298, opened 2016-05-05
Fix PPX for 4.03 (empty let bindings)

vasilisp:
OCaml 4.03 imposes stricter invariants on the parsetree. In particular, empty `let` bindings are not allowed. This PR applies enough fixes to compile `ocsigen-toolkit` (ocsigen/ocsigen-toolkit#17, cc @TobiasBora), but I haven't tested any further, and I am not sure I fixed everything.

### DIFF
--- a/src/ppx/ppx_eliom_client.ml
+++ b/src/ppx/ppx_eliom_client.ml
@@ -89,21 +89,25 @@ module Pass = struct
     ]
 
   let define_client_functions ~loc client_value_datas =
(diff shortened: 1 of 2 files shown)


Drup:
That looks good to me.

vasilisp:
@Drup, thanks. You are faster than Jenkins!

Drup:
Well, it's a short patch. :p

PR #299, opened 2016-05-09
Catch up with TyXML

vasilisp:
Big but totally boring patch. Needs to be merged at the same time as ocsigen/js_of_ocaml#476 and the Ocsigenserver `tyxml` branch.

@balat @vouillon, you will need to do a mass `Html5 -> Html` replacement and use unit arguments for some attributes (`a_defer`,  etc.).


Drup:
@vasilisp Did you added something for https://github.com/ocsigen/tyxml/issues/108 ?

PR #305, opened 2016-05-24
Fix client values in Eliom_service

vasilisp:
Producing client values at the time `Eliom_service` is initialized is problematic; see discussion in 1f0afdfd0144201e71698c35a9c47fd22c10b446. This patch deals with the issue by using a mutable option for the `client_fun` field. The `set_client_fun` function can be used to initialize the `client_fun` later (e.g., by `Eliom_registration`).

The reload function is now represented as `Rf_keep | Rf_client_fun`; the `Rf_client_fun` means that we need to use the `client_fun` for reload. Only services with no POST parameters can have a reload function. `Obj.magic` is no longer needed. (I suspect it was previously hiding problems...)

This is a prefix of the `client-registration` branch implementing client-side `Eliom_registration`, but it is a subtle issue, so I prefer it receives adequate attention in its own.

@balat @vouillon can you take a look?

### DIFF
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -436,6 +436,9 @@ let current_pseudo_fragment = ref ""
 let url_fragment_prefix = "!"
 let url_fragment_prefix_with_sharp = "#!"
 
+let default_reload_function () () =
+  Dom_html.window##location##reload()
+
 let reload_function = ref None
 let reload_functions = ref []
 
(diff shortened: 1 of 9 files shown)


vasilisp:
I have somehow broken actions and reload (which used to work on top of this). Investigatiing...

vasilisp:
OK, the above trouble was related to 9adfdd930bfeb51807486357943b96278d49397b which left us without reload functions in some contexts. About to check with 7e832f7.

Issue #271, opened 2016-02-24
Implement client side actions


balat:
By default Eliom's action services generate a page on server side.
In the case of a mobile app, the page should be generated on client side after action ...

At the moment, it is not recommended the generate pages server-side for client applications, because this would probably send client values (which does not work if the server side app has been updated).

- We should probably have two cases: either the action is performed on server side or there is a client side implementation, in which case it is just a function call (+ redraw) see #272
- *Warning:* Registering an action sets field `send_appl_content` to XAlways on server side. Should be the same on client side! (Eliom_registration.Action)

balat:
I suggest to simplify the semantics, and to have exactly the same behaviour as for other services: actions must be implemented on client side for mobile apps, as all otehr services. If the action is supposed to do soething on the server, just call a server function.

To keep the exact same semantics, we should redisplay the page by default after the action ...

vasilisp:
Done (#308 + #336).

PR #308, opened 2016-05-31
Client-side service registration

### LINKED ISSUE #253: Client version of register functions to replace Eliom_content.set_client_fun
And make possible to register in shared sections

vasilisp:
This PR implements an `Eliom_registration` module whose API is very similar to the server-side `Eliom_registration`. This allows registering client- and server-side implementations of services with largely-shared code.

We provide `Html`, `Action`, `App`, `Redirection`, and `Any` submodules.

The registration functions require an `?app` argument (also added to the server versions for uniformity), which will hopefully go away in the future.

Fixes #253 and #271.

I have made available [an example](https://gist.github.com/vasilisp/0d66a8430697777f4c638457240ad133) demonstrating the programming style that we enable. The services are created on the server and injected to the client, the handlers have implementations on both sides, and the registration is shared.

The low-level way of registering client-side handlers (`Eliom_content.set_client_fun`) still works, so this PR shouldn't involve breaking changes for anybody.


vasilisp:
I am about to re-start modifying `Eliom_registration` (etc.), this time for client-side service identification (routing?). Any objections to merging this?

PR #309, opened 2016-06-03
Add cookie substitutes for iOS WkWebView apps

pwbs:
Context: on iOS WkWebView apps, on the first launch of the app,
cookies don't work at all. After a pause or an exit, they seem to
work at least "OK". But it's not acceptable for an app to be that
broken on the first launch. So, this patch implements a work-around
consisting in using (client) local storage to replace the cookies.
The "cookie substitutes" transit via HTTP headers.

(Implemented with help and advice from Vincent Balat and Jérôme Vouillon.)

modified:   src/lib/client/eliommod_cookies.ml
modified:   src/lib/eliom_client.client.ml
modified:   src/lib/eliom_common.server.ml
modified:   src/lib/eliom_common_base.shared.ml
modified:   src/lib/eliom_process.client.ml
modified:   src/lib/eliom_request.client.ml
modified:   src/lib/server/eliommod_pagegen.ml

### DIFF
--- a/src/lib/client/eliommod_cookies.ml
+++ b/src/lib/client/eliommod_cookies.ml
@@ -25,41 +25,61 @@ include Eliom_cookies_base
 (* CCC The tables are indexed by the hostname, not the port appear.
    there are no particular reason. If needed it is possible to add it *)
 let cookie_tables = Jstable.create ()
-let get_table = function
+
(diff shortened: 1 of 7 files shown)


pwbs:
replaced by https://github.com/ocsigen/eliom/pull/310

PR #310, opened 2016-06-03
Add cookie substitutes for iOS WkWebView apps

pwbs:
Context: on iOS WkWebView apps, on the first launch of the app,
cookies don't work at all. After a pause or an exit, they seem to
work at least "OK". But it's not acceptable for an app to be that
broken on the first launch. So, this patch implements a work-around
consisting in using (client) local storage to replace the cookies.
The "cookie substitutes" transit via HTTP headers.

(Implemented with help and advice from Vincent Balat and Jérôme Vouillon.)

modified:   src/lib/client/eliommod_cookies.ml
modified:   src/lib/eliom_common.server.ml
modified:   src/lib/eliom_common_base.shared.ml
modified:   src/lib/eliom_request.client.ml
modified:   src/lib/server/eliommod_pagegen.ml

### DIFF
--- a/src/lib/eliom_request.client.ml
+++ b/src/lib/eliom_request.client.ml
@@ -115,6 +115,17 @@ let nl_template =
 (* Warning: it must correspond to [nl_template]. *)
 let nl_template_string = "__nl_n_eliom-template.name"
 
+
+
+(** Same as XmlHttpRequest.perform_raw_url, but:
+    - sends tab cookies in an HTTP header
+    - does half and full XHR redirections according to headers
+
+    The optional parameter [~cookies_info] is a pair
+    containing the information (secure, path)
+    that is taken into account for finding tab cookies to send.
+    If not present, the path and protocol are taken from the URL.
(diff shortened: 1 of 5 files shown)


src/lib/eliom_request.client.ml
```diff
@@ -140,6 +151,20 @@ let send
...
       | _ -> [ Eliom_common.tab_cookies_header_name,
                encode_header_value cookies ] in
+    let headers =
+      if Js.Unsafe.global##___eliom_use_cookie_substitutes_ <> Js.undefined then
+        (* Cookie substitutes are for iOS WKWebView *)
+        let host = match host with
+          | None -> "/substitutes" | Some h -> (h  ^ "/substitutes") in
+        let cookies =
+          Eliommod_cookies.get_cookies_to_send
+            ~in_local_storage:true (Some host) https path in
```
vouillon: You should just pass the initial `host` value here, without suffixing it with `/substitutes`

PR #312, opened 2016-06-12
Do not use deprecated Xml_print modules

vasilisp:
This PR eliminates use of the `Xml_print` modules that are deprecated (as of TyXML 4.0).

This has an impact on the printers that we export, so it is a breaking change.

@Drup, this is well within your jurisdiction, and I am not very proficient in `Format`, so please take a look :).

### DIFF
--- a/src/lib/eliom_registration.server.ml
+++ b/src/lib/eliom_registration.server.ml
@@ -107,29 +107,26 @@ module Make_typed_xml_registration
   (Typed_xml: Xml_sigs.Typed_xml with module Xml := Xml)
   (E : sig type content end) = struct
 
-    module Format = Xml_print.Make_typed(Xml)(Typed_xml)(Ocsigen_stream.StringStream)
+    module Fmt = Xml_print.Make_typed_fmt(Xml)(Typed_xml)
 
-    let result_of_content_subxhtml get_etag c =
+    let out =
       let encode x = fst (Xml_print.Utf8.normalize_html x) in
-      let x = Format.print_list ~encode c in
+      Format.asprintf "%a" (Fmt.pp_elt ~encode ())
+
+    let out_list l =
+      List.map out l
+      |> String.concat ""
+      |> Ocsigen_stream.StringStream.put
+      |> Ocsigen_stream.StringStream.make
+
+    let result_of_content c =
       let default_result = Ocsigen_http_frame.Result.default () in
       Lwt.return
         (Ocsigen_http_frame.Result.update default_result
           ~content_length:None
           ~content_type:(Some "text/html")
-          ~etag:(get_etag c)
           ~headers:(Http_headers.dyn_headers)
-          ~stream:(x, None) ())
-
-    module Cont_content =
-      (* Pasted from ocsigen_senders.ml and modified *)
-      struct
-
-        let get_etag ?options c = None
-
-        let result_of_content c = result_of_content_subxhtml get_etag c
-
-      end
+          ~stream:(out_list c, None) ())
 
     module Cont_reg_base = struct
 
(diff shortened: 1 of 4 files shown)


src/lib/server/extensions/atom_feed.ml
```diff
@@ -131,19 +131,19 @@ let rec metaAttr_extract l = match l with
...
 r
 
-let print_html5 l =
+let print_html5 e =
   let buffer = Buffer.create 500 in
-  let output = Buffer.add_string buffer in
   let encode x = fst (Xml_print.Utf8.normalize_html x) in
-  Eliom_content.Html.Printer.print_list ~encode ~output l;
+  Buffer.add_string buffer
+    (Format.asprintf "%a" (Eliom_content.Html.Printer.pp_elt ~encode ()) e);
```
Drup: Use `Format.formatter_of_buffer` to make a formatter out with the buffer, and then `(pp_elt ~encode ()) fmt e`.

src/lib/eliom_registration.server.ml
```diff
@@ -107,29 +107,26 @@ module Make_typed_xml_registration
...
+    let out =
       let encode x = fst (Xml_print.Utf8.normalize_html x) in
-      let x = Format.print_list ~encode c in
+      Format.asprintf "%a" (Fmt.pp_elt ~encode ())
+
+    let out_list l =
+      List.map out l
+      |> String.concat ""
+      |> Ocsigen_stream.StringStream.put
+      |> Ocsigen_stream.StringStream.make
```
Drup: `Format.pp_print_list ~sep:(fun _ () -> ())` to output lists.

Morally, we should get rid of StringStream and provide a buffer-like API (which would make creation of a formatter easy), but in the meantime, you can probably make a formatter out of the put function (or just don't bother for now).

src/lib/server/extensions/eliom_atom.ml
```diff
@@ -33,12 +33,14 @@ module Atom_info = struct
...
-module Format = Xml_print.Make_simple(Tyxml_xml)(Atom_info)
+module Fmt = Xml_print.Make_fmt(Tyxml_xml)(Atom_info)
 
 let result_of_content feed headers =
   let b = Buffer.create 10 in
   let encode x = fst (Xml_print.Utf8.normalize_html x) in
-  Format.print_list ~output:(Buffer.add_string b) ~encode [Atom_feed.xml_of_feed feed];
+  (Atom_feed.xml_of_feed feed
+   |> Format.asprintf "%a" (Fmt.pp ~encode ())
+   |> Buffer.add_string b);
```
Drup: Why do we need a buffer here ? We just return the string.

src/lib/server/extensions/atom_feed.ml
Drup: Actually, the buffer doesn't seem to be used. We can just return the string.

vasilisp: Thanks, I will remove the buffers , I was blindly adapting what was there before.

src/lib/eliom_registration.server.ml
```diff
@@ -107,29 +107,25 @@ module Make_typed_xml_registration
...
-
-    let result_of_content_subxhtml get_etag c =
-      let encode x = fst (Xml_print.Utf8.normalize_html x) in
-      let x = Format.print_list ~encode c in
+    module Fmt = Xml_print.Make_typed_fmt(Xml)(Typed_xml)
+
+    let out l =
+      Format.asprintf "%a"
+        (Format.pp_print_list ~pp_sep:(fun _ () -> ())
+           (let encode x = fst (Xml_print.Utf8.normalize_html x) in
```
Drup: It would probably be best to hoist that.

Also, you probably want to check the first version with pp_print_list, we may need to reimplement it locally.

PR #331, opened 2016-07-04
Eliom_notif

canonici:
I'm not sure if I put the files in the right place.

### DIFF
--- a/src/lib/eliom_notif.server.ml
+++ b/src/lib/eliom_notif.server.ml
@@ -1,0 +1,190 @@
+open Lwt
+
+module type S = sig
+  type identity
+  type key
+  type notification
+  val equal_key                  : key -> key -> bool
+  val equal_identity             : identity -> identity -> bool
+  val get_identity               : unit -> identity Lwt.t
+  val max_resource               : int
+  val max_identity_per_resource  : int
+end
+
+module Make (A : S) = struct
+
+  type notification_data = A.key * A.notification
+
+  type notification_react =
+    notification_data Eliom_react.Down.t
+    * (?step: React.step -> notification_data -> unit)
+
+  module Notif_hashtbl = Hashtbl.Make(struct
+    type t    = A.key
+    let equal = A.equal_key
+    let hash  = Hashtbl.hash
+  end)
+
+  module Weak_tbl = Weak.Make (struct
+    type t = (A.identity * notification_react) option
+    let equal a b = match a, b with
+      | None, None ->
+	true
+      | Some (a, b), Some (c, d) ->
+	A.equal_identity a c && b == d
+      | _ -> false
+    let hash = Hashtbl.hash
+  end)
+
+  module I = struct
+
+    let tbl = Notif_hashtbl.create A.max_resource
+      
+    let lock = Lwt_mutex.create ()
+
+    let async_locked f = Lwt.async (fun () ->
+      Lwt_mutex.lock lock >>= fun () ->
+      f ();
+      Lwt.return (Lwt_mutex.unlock lock)
+    )
+
+    let remove_if_empty wt key = async_locked (fun () ->
+      if Weak_tbl.count wt = 0
+      then Notif_hashtbl.remove tbl key
+    )
+
+    let remove v key = async_locked (fun () ->
+      let () =
+	try
+	  let wt = Notif_hashtbl.find tbl key in
+          Weak_tbl.remove wt v;
+	  remove_if_empty wt key
+	with Not_found -> ()
+      in
+      Lwt.return ()
+    )
+
+    let add v key = async_locked (fun () ->
+      let wt =
+	try
+	  Notif_hashtbl.find tbl key
+        with Not_found ->
+	  let wt = Weak_tbl.create A.max_identity_per_resource in
+          Notif_hashtbl.add tbl key wt;
+          wt
+      in
+      if not (Weak_tbl.mem wt v)
+      then Weak_tbl.add wt v;
+      Lwt.return ()
+    )
+
+    let iter =
+      let iter (f : Weak_tbl.data -> unit Lwt.t) wt : unit =
+	Weak_tbl.iter
+	  (fun data -> Lwt.async (fun () -> f data))
+	  wt
+      in
+      fun f key -> async_locked (fun () ->
+	let () =
+	  try
+	    let wt = Notif_hashtbl.find tbl key in
+	    let g data = match data with
+              | None ->
+		Weak_tbl.remove wt data;
+		remove_if_empty wt key;
+		Lwt.return ()
+              | Some v ->
+		f v;
+		Lwt.return ()
+	    in
+            iter g wt;
+	  with Not_found -> ()
+	in
+	Lwt.return ()
+      )
+  end
+
+  let identity_r : (A.identity * notification_react) option Eliom_reference.eref =
+    Eliom_reference.eref
+      ~scope:Eliom_common.default_process_scope
+      None
+
+  (* notif_e consists in a server side react event,
+     its client side counterpart,
+     and the server side function to trigger it. *)
+  let notif_e : notification_react option Eliom_reference.eref =
+    Eliom_reference.eref
+      ~scope:Eliom_common.default_process_scope
+      None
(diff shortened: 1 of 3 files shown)


src/lib/server/eliom_notif.mli
```diff
@@ -0,0 +1,62 @@
...
+	      ) : unit)
+           ]
+  
+  *)
+
+  val client_ev : unit -> (A.key * A.notification) Eliom_react.Down.t
+
+  val unlisten_wrapper : key:A.key -> handler:'a  -> 'a
+
+  val listen_wrapper : key:A.key -> handler:'a  -> 'a
```
Drup: You probably don't want to show those in the documentation (use `(**/**)`).

src/lib/server/eliom_notif.mli
```diff
@@ -0,0 +1,62 @@
...
+      to notifications from the server.
+      For example:
+
+      let%client handle_notification some_stuff ev =
+         ...
+
+      let%server something some_stuff =
+         ignore
+           [%client
+              (ignore (React.E.map
```
Drup: `Lwt_react.E.keep` ?

src/lib/server/eliom_notif.ml
```diff
@@ -0,0 +1,180 @@
...
+      (fun () ->
+         let e, send_e = React.E.create () in
+         let client_ev = Eliom_react.Down.of_react
+             (*VVV If we add throttling, some events may be lost
+               even if buffer size is not 1 :O *)
+             ~size: 100 (*VVV ? *)
+             ~scope:Eliom_common.default_process_scope e in
+         (client_ev, e, send_e)
+         (* I don't really need e, but I need to keep a reference on it during
+            the session to avoid it being garbage collected. *))
```
Drup: Isn't `client_ev` going to keep a reference anyway ?

Drup:
It's not clear to me how to instantiate this with the very simple notion of identity that is included in eliom (each client is an identity), and then how to use it. I think we need a complete example. :]

Also, I'm not sure how I like the fact that the functor creates new tables.

vasilisp:
Good job!

I would prefer `src/lib/eliom_notif.server.ml` and `src/lib/eliom_notif.server.mli`. The `server` and `client` directories are reserved for lower-level stuff (`eliommod`).

To expand on @Drup's point on identities, we need to document the contexts where the various functions are allowed (during a request?). From the type of `get_identity`, it is not clear at all how to provide an identity, given that the unit argument doesn't say anything. So the calling context (of `listen` et al., and by transitivity of `get_identity`) is very crucial.

canonici:
I removed the reference to e, it looks like it was not needed.
I did an example (module MailStorage is pretty bad but it is only for the demo...).
[demo code](https://gist.github.com/canonici/6db5b2d818a9424110103211a8a4c5a7)

Drup:
```
let get_identity () = Eliom_reference.get username |> Lwt_main.run
```

If that's not a sign that we need `get_identity` to return a thread ....

Drup:
> module MailStorage is pretty bad but it is only for the demo...

Well, no, that's not how it works. We need to improve the API until both the demo and the advance examples looks good :)

Drup:
I see lot's of uses of the polymorphic comparison `(=)` both in the implementation and in the example. As a general rule, Please don't. polymorphic comparison should never be used for any serious task. You use ppx deriving in the example, just derive equal.

Other than that, I think this mail example should be turned into a nice tutorial. :3

canonici:
At first I had get_identity return a thread, but then I was somehow tricked into returning a value of type identity by Eba_current_user.Opt.get_current_userid, my bad.

I'm not sure as to why polymorphic comparison should never be used?
I know it can have troubles with things like:
`let rec l = 1 :: l in l = l`

But is there any danger with primitive types?

Drup:
Polymorphic comparison is slow if it's not specialized (and it was not, in your examples). Furthermore, if equals is used on functional values (like ... objects), it's going to explode in your face.
It's better now.

canonici:
Ok, I had no idea, thanks!

canonici:
I added a mechanism to clean the hashtable from empty elements but it is currently arbitrarily set to run every 60 seconds and it iterates on the whole table, if someone has a better solution I'll take it.

Do we want to let people set their own frequency so they can eventually disable it?

src/lib/eliom_notif.server.ml
```diff
@@ -0,0 +1,197 @@
+open Lwt
+
+module type S = sig
+  type identity
+  type key
+  type notification
+  val equal_key                  : key -> key -> bool
+  val equal_identity             : identity -> identity -> bool
+  val get_identity               : unit -> identity Lwt.t
+  val max_ressource              : int
```
vasilisp: It is spelled `resource` in English :).

src/lib/eliom_notif.server.mli
```diff
@@ -0,0 +1,59 @@
...
+module type S = sig
+  type identity
+  type key
+  type notification
+  val equal_key                  : key -> key -> bool
+  val equal_identity             : identity -> identity -> bool
+  val get_identity               : unit -> identity Lwt.t
+  val max_ressource              : int
+  val max_identity_per_ressource : int
+end
```
vasilisp: Can you document this signature? Especially the `max_*` values.

src/lib/eliom_notif.server.ml
```diff
@@ -0,0 +1,197 @@
...
+      else
+        content_gen identity >>= fun content -> match content with
+        | Some content -> send_e (key, content); Lwt.return ()
+        | None -> Lwt.return ()
+    in
+    (* on all tabs registered on this data *)
+    I.iter f key
+
+  let client_ev () =
+    let (ev, _) = Lwt_main.run (
```
vasilisp: I am uncomfortable with this `Lwt_main.run`. Can `client_ev` return an Lwt thread instead?

vasilisp:
I don't know of a good solution for your mini garbage collector. Maybe export a call for this and let the user trigger it? I wouldn't expect an OCaml module to wake up on its own and do computation.

src/lib/eliom_notif.server.mli
```diff
@@ -0,0 +1,84 @@
...
+    [max_resource] is the initial size for the hash table storing the data of
+    clients listenning on resources, for best results it should be on the order
+    of the expected number of different resources one may want to be able to
+    listen to.
+    [max_identity_per_resource] is the initial size for the tables storing the
+    data of clients listenning on one given resource, fo best results it
+    should be on the order of the expected number of clients that may listen
+    on a given resource.
+*)
+
```
vasilisp: Can you document the fields with individual comments attached to them?

src/lib/eliom_notif.server.mli
```diff
@@ -0,0 +1,84 @@
...
+         ignore
+           [%client
+              (ignore (React.E.map
+		        (handle_notification ~%some_stuff)
+		        ~%(Notif_module.client_ev ())
+	      ) : unit)
+           ]
+  
+  **)
+
```
vasilisp: Don't leave a blank line between the documentation comment and the thing documented, ocamldoc mandates that they are consecutive.

src/lib/eliom_notif.server.mli
```diff
@@ -0,0 +1,84 @@
...
+  val client_ev : unit -> (A.key * A.notification) Eliom_react.Down.t Lwt.t
+
+
+  (** Call [clean freq] to launch an asynchronous thread clearing the tables
+      from empty data every [freq] seconds
+  *)
+  val clean : float -> unit Lwt.t
```
vasilisp: I was actually thinking of a `clean` operation that does a cleanup only once, not a looping one (so that the client will be responsible for doing it when they deem worthwhile). But I am not sure on this :).

agarwal:
The mli file has no information about what this new module is about. I think every mli should have a paragraph or more explaining the purpose of the module.

Also, the module type name `S` is almost always used to mean the return type of a functor, not the input type. I would pick a different name for the input type, ideally a word that captures the semantics of the input. If that is too difficult, then I'd go with `Arg` or maybe `A`.

PR #333, opened 2016-07-06
Bad submodule name and more descriptions

dannywillems:


doc/manual-wiki/server-services.wiki
```diff
@@ -35,8 +35,8 @@ The service creation could be split in two steps:
   <<a_api | val Eliom_service.App.service >>, or
   <<a_api | val Eliom_service.OCaml.service >>
   (depending on the type of service you want to create); and
-* register a service handler, e.g., using <<a_api | val
```
vasilisp: Nitpicking, but `e.g.` should be followed by a comma.

doc/manual-wiki/server-services.wiki
```diff
@@ -35,8 +35,8 @@ The service creation could be split in two steps:
...
-  Eliom_registration.HTML5.register >>.
```
vasilisp: That has all changed in `master`. You would use `Eliom_service.create` (there is no `Html5` submodule) and `Eliom_registration.Html.register`. Given that the manual is tied to an Eliom version, there is little point in fixing the doc for the old API.

PR #335, opened 2016-07-11
eliom-distillery: add list-template to list templates

dannywillems:
It could be useful if later more templates are added.


vasilisp:
Can you pluralize it (`-list-templates`)? Other than that, looks good.

dannywillems:
@vasilisp done.

PR #336, opened 2016-07-19
Client routing

vasilisp:
This PR introduces client-side service routing, and associated improvements to client-side service registration.

The routing code is now shared between server and client. What used to be `Eliommod_services` (on the server) is now `Eliom_route_base` (shared). I functorized the code (`Eliom_route_base.Make`) in the process of discovering what needs to be implemented on the client. The functor interface is far from pretty, but it is only for internal use. The client implements some of the required functionality (e.g., doubly-linked lists used for garbage collection) only in a dummy manner.

The entry points for client-side routing are `Eliom_client.change_page_unknown` and `Eliom_client.change_page_uri`.


balat:
Ok thank you!

I had a look but it is difficult to check everything.
Besport now works well with that version.
I suggest we merge this if you think we can (to avoid maintaining two branches of eba bs, etc)
 once you've done the change we discussed about the interface (change_page_uri/unknown).

vasilisp:
I am doing the API changes discussed (you will need to remove the `~client` argument on your side), and on a small fix for something I broke. We can merge later today.

vasilisp:
@balat , we are ready to merge.

PR #338, opened 2016-07-21
fix tiny bug leading to wrong and confusing error messages

jrochel:


vasilisp:
Can you provide an example of a wrong message? It seems to me that these options are also meant for the `eliom` tag, not just for the `eliommodule` tag`. If we have to pick one of the two for the error messages, I would go with `eliom`.

vouillon:
Indeed, these options are meant for both tags. So, the right fix would be to pass the tag name around.

jrochel:
Better?

vasilisp:
I think it's OK now, merging. Thanks!

PR #340, opened 2016-07-21
Documentation for multiple functions: description of the last parameters

dannywillems:
Same documentation than last PR for other functions. Deduction based on the parameters names in ml file. Had to be in one PR but you were too fast.


Drup:
As a general documentation rules, writing "The nth parameter does X" is quite bad. A better way to write it is:
"[change_page ~service get post] does blabla. [get] are the GET parameters and [post] are the POST parameters".

dannywillems:
I agree with that but I don't have the description about all parameters, so I used this way. Do you want I change the syntax with yours without describing other parameters?

Drup:
Well, if you can't figure out what all the parameters mean, then we have a big issue that we should fix anyway, so yes, I think it would be better to change the description.

Issue #348, opened 2016-07-27
Use PPX internally


vasilisp:
I think all the prerequisites are there: we have PPX extensions for JSOO and Lwt, we have `ppx_deriving_json`, and we have killed `pa_include`.

There are no backwards-compatibility issues. We have broken compatibility with OCaml 4.01 anyway, and we can still provide the Camlp4 extension for external code that still needs it.

Using PPX will possibly speed up compilation, we will be able to develop Eliom in saner language, and it will hopefully help @Drup :).

vasilisp:
The branch `ppx-internally` seems to work. CC @Drup .

I did a pass through `camlp4-to-ppx`, fixed some Camlp4-isms, and made some changes for `ppx_deriving_json` (whose output is not identical to the Camlp4 version).

We also need a small fix in `ppx_deriving_json`: ocsigen/js_of_ocaml@0a7fea9724b5c4c74a30ab774d65c8d8a9cfea7e .

Drup:
Hum, apparently we still have a necessary dependency to camlp4 for optcomp ... Can we get rid of that too and use cppo instead ? :]

vasilisp:
Yes, I haven't dealt with the tools yet, just with the lib.
 
Sure, we will get rid of `optcomp` :).

vasilisp:
Actually, `optcomp` had become useless when we dropped support for 4.01 and older. So I dropped it.

We can't yet make the dependency on Camlp4 optional because JSOO uses it.

vasilisp:
On my laptop, compilation time goes from about 3.5 minutes to about 2 minutes (and we could probably do better with `ppx_driver` and less forking).

vasilisp:
Fixed by #349 .

PR #349, opened 2016-07-28
Use PPX internally

vasilisp:
See #348 for discussion.

I haven't noticed any ill-effects. Let's see what Jenkins says.


vasilisp:
If anybody wants to review, the first commit is mechanical translation, so skim through it or ignore it. The rest should be readable.

vasilisp:
@vouillon would merging this cause any trouble for BeSport? Normally it shouldn't, and no changes are needed, but you may want to check.

PR #351, opened 2016-07-29
The handler for beforeunload events should return undefined rather than null

vouillon:
This is what is expected by internet explorer.


vasilisp:
Fine, as long as Firefox/Chrome/WebKit are still happy :). They don't quite agree on what `onbeforeunload` should return and how, e.g., WebKit wants us to return a string (c013ff3190277a9e4b34981a5a0228021972c9c1), while some of the others want us to set the `returnValue` field of the event.

vouillon:
I don't see why they would not be happy. Returning `undefined` is like not returning anything at all. Returning `null` is weird here.

PR #352, opened 2016-07-29
Separate onunload and onbeforeunload callbacks

vouillon:
We want to have `onbeforeunload` callbacks that are able to cancel page changes, and `onunload` callbacks that are invoked when we are sure that the page is going to change (and may have destructive effects).

This changes the type of `onunload` callbacks from `unit -> string option` to `unit -> _` to avoid breaking existing code. Eventually, type `unit -> unit` would be better.

### DIFF
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -103,16 +103,10 @@ let _ =
 
 let onunload_fun _ =
   update_state ();
-  (* running remaining callbacks, if onbeforeunload left some *)
-  let _ = run_onunload ~final:true () in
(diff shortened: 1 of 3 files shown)


Drup:
We already breaking a good amount of user code with this version, we might as well do this small one too ...

vasilisp:
I agree with @Drup, let's just break the API. In fact, returning a `string option` won't have the effect it had in Eliom 5.0, so we would be causing runtime breakage. I much prefer compile-time breakage :).

vasilisp:
Looks good, so I take the liberty of pushing the green button so that we can move on.

PR #359, opened 2016-08-11
Split-out service functions

### LINKED ISSUE #311: Eliom6 introduces some redundancy in declaration of attached post co-services
Consider the minimal example of an attached POST co-service listed at the end.  It implements a simple 2-step service which presents a form on the 1st step and displays the submitted values on the 2nd step.

The declaration of the (GET, POST) parameters for the 2nd step is `Post (Eliom_parameter.unit, Eliom_parameter.(int "n" ** string "s"))`.  However, the GET parameter declaration is redundant, because we know already it must be `Eliom_parameter.unit` from the fallback.  Note that previous versions of Eliom did not have this redundancy.

```ocaml
open Eliom_content.Html.F

let coucou_service =
    Eliom_service.create
        ~meth:(Get Eliom_parameter.unit)
        ~id:(Path [])
        ()

let coucou_handler () () =
    let rec step1_handler () () =
        let step2_service =
            Eliom_registration.Html.create
                ~scope:Eliom_common.default_session_scope
                ~meth:(Post (Eliom_parameter.unit, Eliom_parameter.(int "n" ** string "s")))
                ~id:(Fallback coucou_service)
                step2_handler in
        let step2_form (n, s) =
            [
            Form.input ~input_type:`Text ~name:n ~value:0 Form.int;
            Form.input ~input_type:`Text ~name:s ~value:"hello" Form.string;
            Form.button_no_value ~button_type:`Submit [pcdata "Submit"];
            ] in
        Lwt.return
            (html
            (head (title (pcdata "Main")) [])
            (body [Form.post_form step2_service step2_form ()]))
    and step2_handler () (n, s) =
        Lwt.return
            (html
            (head (title (pcdata "Step 2/2")) [])
            (body
                [
                p [pcdata ("N=" ^ string_of_int n)];
                p [pcdata ("S=" ^ s)];
                ])) in
    step1_handler () ()

let () =
    Eliom_registration.Html.register coucou_service coucou_handler
```

### LINKED ISSUE #322: Attached co-services with non-unit GET params in Eliom 6
Perhaps I'm missing something, but the definition of the type `Eliom_service.id` seems to preclude fallback services with non-unit GET params.  Is this intentional?  If so, isn't the redeclaration of GET parameters for the co-service redundant, since they have to be `Eliom_parameter.unit`?

In any case, fallbacks with non-unit GET params would be useful.  Here's one minimal example:

````ocaml
open Eliom_content.Html.F

let coucou_service = Eliom_service.create
    ~meth:(Get Eliom_parameter.(int "x"))
    ~id:(Path [])
    ()

let rec step1_handler x () =
    let step2_service =
        Eliom_registration.Html.create
            ~scope:Eliom_common.default_session_scope
            ~meth:(Post (Eliom_parameter.int "x", Eliom_parameter.string "s"))
            ~id:(Fallback coucou_service)
            step2_handler in
    let step2_form s =
        [
        Form.input ~input_type:`Text ~name:s ~value:"hello" Form.string;
        Form.button_no_value ~button_type:`Submit [pcdata "Submit"];
        ] in
    Lwt.return
        (html
        (head (title (pcdata "Main")) [])
        (body [Form.post_form step2_service step2_form x]))

and step2_handler x s =
    Lwt.return
        (html
        (head (title (pcdata "Step 2/2")) [])
        (body
            [
            p [pcdata ("X=" ^ string_of_int x)];
            p [pcdata ("S=" ^ s)];
            ]))

let () =
    Eliom_registration.Html.register coucou_service step1_handler
````

vasilisp:
This PR separates-out some cases previously handled by the unified `Eliom_service.create` function:

- `Eliom_service.create_external`
- `Eliom_service.attach_get`
- `Eliom_service.attach_post`

Closes #311 .
Closes #322 .

The `~id` argument of `Eliom_service.create` has been renamed to `path`, and it is to be thought of as a "typed" option with constructors `Path` and `No_path`. I think retiring the `Global` name will help documentation-wise.

Additionally, the `PUT` and `DELETE` verbs get back their `raw_post_data` parameter. There is currently no way of creating attached `PUT` and `DELETE` coservices.


vasilisp:
Providing `attach_get` and `attach_post` services everywhere in `Eliom_registration` requires an amount of boilerplate, which is unfortunate. At least it is way less boilerplate than we removed with #293 .

Drup:
I'm still quite sad that we have to split attach get/post. It's also disappointing that we can only shave one type variable off meth/service. I vote to shorten `create_external` in simply `external`.

vasilisp:
`attach_get` and `attach_post` are barely related in what they do to the parameters. We can encode the differences in another GADT, but I don't think it's worth it.

vasilisp:
`external` is a keyword.

Drup:
Doh. :(

Maybe `extern` ? I don't really like `create_*` ...

balat:
The advantage of the new interface is:
* It has all the power of the old service model (coservices on top of services with get parameters) the only exception being attached coervices with put/delete (but do we need this?)
* It is much simpler than the old interface and even than the current dev version. Nobody can say that the service model is complex now ;)

The drawback is that attached coservices have a different interface... :'(
As some kind of consolation, they are, I think, less used than the other services (non-attached coservices and regular services).

About `create_external`: we already have `create` so `create_external` makes sense for me. But ok for `extern`.

About `attach_get`: may be `attached_get`? We already have a function `attach` for the action of attaching a coservice to a service. Here we are not attaching but creating a attached service.

The consequence of the new interface is to change the vocabulary.
We will probably remove the term "coservice" completely from the documentation.
The new interface for `create` is simple enough to be understood, but I'm still not completely happy for `attach(ed)_get` and `attach(ed)_post`. But I have no better idea for now.

Last remarks, about `attach_global_to_fallback`: Do we still want to use the term global? There is no contructor called `Global` anymore if I'm right. If we keep this function name, we must at least explain what a global service is in the API doc and manual.

Also the API doc is not up to date (still mentioning [Global], [Fallback], etc.).

vasilisp:
> The drawback is that attached coservices have a different interface... :'(

I have personally been persuaded that they need a separate interface. They substantially differ from the other services, as they establish a relationship with the parameters of a preexisting service (possibly contradicting the explicitly specified parameters). Accomodating this requires type system gymnastics and eventually leads to a harder-to-use interface.

> About `attach_get`: may be `attached_get`?
> ...
> Last remarks, about `attach_global_to_fallback`

I have no strong opinions on the names. I think that `attach_global_to_fallback` needs to be renamed. Maybe `attach_preexisting`? The `preexisting` designation makes `attach_{get,post}` more acceptable, by implying that "by default" we attach something that we create at the same time.

> Also the API doc is not up to date (still mentioning [Global], [Fallback], etc.).

Let's agree on the API first. I will fix the doc before merging the PR.

vasilisp:
The documentation is mostly up-to-date with the new API. I will make adjustments after the merge because it is hard to predict the output on ocsigen.org .

There are `service-split-att` compatibility branches for `ocsigen-start` et al.

There is a `service-split-att` branch for the `tutorial` repo which updates everything for the new API.

CC @balat @vouillon @dannywillems

paurkedal:
Wouldn't it be better to treat PUT and DELETE similar to POST with respect to parameters?  That is, they both receive decoded instead of raw data?  Especially PUT and POST requests are similar as they could take exactly the same POST data: The PUT request would create/replace an item named by the GET part, whereas the POST would create an item with a generated name under the collection named by the GET part.

PR #367, opened 2016-08-25
Remove template 'mobile'. Too old

dannywillems:
The directory with the name "mobile" still present in `$HOME/.opam/$COMPILER/lib/eliom/templates` so `eliom-distillery -list-templates` still list it (in distillery.ml, the templates list is generated with `Findlib.package_directory "eliom.templates").

It could be useful to remove all directories in relation with eliom when using `opam remove eliom`.


vasilisp:
Thanks.

PR #369, opened 2016-08-25
Distillery: check if the project name is valid for the given template…

dannywillems:
See https://github.com/ocsigen/ocsigen-start/issues/102.

It is supposed all templates have a file named forbidden_project_name (can be changed) with the list of forbidden project names.
With this idea of a file in the template listing all forbidden project names, it will be scalable if more templates are done.

I find the code ugly but does the job.

### DIFF
--- a/src/tools/distillery.ml
+++ b/src/tools/distillery.ml
@@ -1,5 +1,6 @@
-
 open Printf
+
+let forbidden_project_name_filename = "forbidden_project_name"
 
 let distillery_basic = "basic.ppx"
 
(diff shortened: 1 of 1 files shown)


src/tools/distillery.ml
```diff
@@ -1,6 +1,7 @@
-
 open Printf
 
+let forbidden_project_name_filename = "forbidden_project_name"
```
vasilisp: I would go with something much shorter, like `reserve` (maybe even a hidden `.reserve` file).

src/tools/distillery.ml
```diff
@@ -194,6 +195,32 @@ let get_templates () =
...
+  let file = Filename.concat
+    (Filename.concat (get_templatedir ()) template_name)
+    forbidden_project_name_filename
+  in
+  if Sys.file_exists file
+  then
+  (
+    let in_file = open_in file in
+    try
+      let rec locale () =
```
vasilisp: The name is confusing to non-french speakers; locale in computing generally points to internationalization :). Maybe `aux` or just `f`?

src/tools/distillery.ml
```diff
@@ -194,6 +195,32 @@ let get_templates () =
...
+        if project_name = (input_line in_file) then false else locale ()
+      in
+      locale ()
+    with
+    | End_of_file -> close_in in_file; true
+    | e           -> raise e
```
vasilisp: It is unnecessary to catch `e` if all you want to do is throw it again.

src/tools/distillery.ml
```diff
@@ -246,8 +273,14 @@ let main () =
           template, name, dir
       | _ -> Arg.usage spec usage_msg; exit 1
     in
-    let env, preds, source_dir = init_project template name in
-    create_project ~name ~env ~preds ~source_dir ~destination_dir:destination_dir
+    if not (check_forbidden_project_name name template) then
```
vasilisp: I would negate the output of `check_forbidden_project_name`. You should return `true` if what you check for (if the project name is forbidden) is true, no?

vasilisp:
Other than the style-related comments above, looks good.

dannywillems:
@vasilisp Done.

vasilisp:
Thanks, looks good.

Issue #368, opened 2016-08-25
[FEATURE] Write templates and share them with the community.


dannywillems:
Is there a way for the moment to allow developer to write their own Eliom templates, let other developers to download them and use them?
It could be very interesting and a powerful feature for Eliom to grow.

Here an idea how it could be organized:

### To write a template.

To create a mobile template for a chat: 
* create a git repository (on GitHub for example) named *chat-mobile*.
* add files
* push on the repository *https://github.com/user/eliom-template-chat-mobile*

### To use the template *chat-mobile*.
We can think a new argument added to eliom-distillery like
```
eliom-distillery -add-template chat-mobile https://github.com/user/eliom-template-chat-mobile
```
This command will create a directory *chat-mobile* in `$HOME/.opam/$COMPILER/lib/eliom/templates`, download and copy the files from the repository https://github.com/user/eliom-template-chat-mobile.
The list-templates argument will give now
```
basic.ppx
basic.camlp4
chat-mobile
```
and we can now use the template with
```
eliom-distillery -name chat-mobile-test -template chat-mobile
```

Drup:
Isn't that already possible ? ~~eba~~ ocsigen-start installs a new template.

Currently, you have to manually put it in the right directory, though.
As a general thing, we could have eliom-distillery be a standalone project (or investigate https://github.com/dbuenzli/carcass).

dannywillems:
@Drup: Yes it's possible but it's not very easy for a beginner to know it. Eliom-distillery as a standalone project can be very interesting.

I implemented a very basic version of `-add-template` and `-rm-template` using git. I must be improved so I don't make a PR for the moment.
Currently, I find it's a very good solution because I can work on different templates at the same time and start with feature I want (I don't need to start a new project with the entire list of Ocsigen-start features).

I think templates must be in seperated directories as Eliom tools.
I think Eliom-distillery will be the first thing new user will see (because if you want to use ocsigen-start, you need to use Eliom-distillery) so improve it with this kind of features could be very interesting. It will help to grow the community and help to begin new Eliom projects from scratch.

dannywillems:
When this feature will be implemented (maybe #413), documentation about `.eliomignore` and `.reserve` must be added.

PR #374, opened 2016-09-16
Add .eliomignore support to not copy some files from a template.

dannywillems:
Due to [this PR on Ocsigen-start](https://github.com/ocsigen/ocsigen-start/pull/138) about an update script from an older version to a newer, we need to have a way to ignore some files while copying. (It doesn't make sense to copy the ~~`update.sql`~~ `upgrade.sql` file in the template).

This PR allows to add a `.eliomignore` file in templates to avoid to copy some file (like `.gitignore`).
It will be more interesting when we will add a feature like #368 (I'm on it, but I don't like the way I implemented it, I need to improve it).

It implements a function `lines_of_files` returning all lines in a file as a string list to be able to check if a filename is in the `.eliomignore` file without doing multiple I/O operations on `.eliomignore`.

I also changed the implementation for `.reserve` by using `lines_of_files` because it did too much I/O operations.

### DIFF
--- a/src/tools/distillery.ml
+++ b/src/tools/distillery.ml
@@ -1,12 +1,12 @@
 open Printf
 
 (* File containing the reserve project names for templates *)
-let reserve_project_name_filename = ".reserve"
+let reserve_project_name_filename = ".eliomreserve"
 
 (* File containing files which must be ignored while copying a template. *)
 let eliomignore_filename          = ".eliomignore"
 
-let distillery_basic = "basic.ppx"
+let distillery_basic              = "basic.ppx"
 
 (* Returns all lines of [file] as a string list. Returns an empty list if [file]
  * doesn't exist.
--- a/src/tools/distillery.ml
+++ b/src/tools/distillery.ml
@@ -1,11 +1,27 @@
 open Printf
 
-(* Constant use for the filename containing the reserve project name for
- * templates
- *)
+(* File containing the reserve project names for templates *)
 let reserve_project_name_filename = ".reserve"
 
+(* File containing files which must be ignored while copying a template. *)
+let eliomignore_filename          = ".eliomignore"
+
 let distillery_basic = "basic.ppx"
+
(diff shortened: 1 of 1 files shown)


src/tools/distillery.ml
```diff
@@ -1,12 +1,28 @@
 open Printf
 
-(* Constant use for the filename containing the reserve project name for
- * templates
- *)
+(* File containing the reserve project names for templates *)
 let reserve_project_name_filename = ".reserve"
 
+(* File containing files which must be ignored while copying a template. *)
+let eliomignore_filename          = ".eliomignore"
```
vasilisp: Just `.reserve` maybe? The `eliom` prefix looks redundant and inconsistent with `.reserve`.

vasilisp:
Looks good other than the name :).

src/tools/distillery.ml
dannywillems: `.reserve` and `.eliomignore` have two different meanings.
- `.reserve` is for the forbidden projects names, which depends on the template.
- `.eliomignore` is for the files which must be ignored while copying the template.

dannywillems: Oh, or you mean `.ignore`?

vasilisp: Yes, just `.ignore`, sorry.

dannywillems: I chose `.eliomignore` because I don't know if some projects use `.ignore`. I mean, for git is `.gitignore` and I suppose there are also other projects which use ignore files and I think is not a good idea to impose `.ignore` for Eliom.

Or maybe replace `.reserve` by `.eliomreserve`.

vasilisp: I much doubt that you would want to put another `.ignore` in an `eliom-distillery` template (we are not striving for ultimate generality), but in any case, I am fine with `.eliomreserve` and `.eliomignore`. My primary concern is consistency.

PR #376, opened 2016-09-16
Remove eliom_loader which is part of ocsigen-start

dannywillems:
Why eliom_loader.ml was in Eliom?


vasilisp:
Thanks. I originally added it to Eliom thinking that we may want to use the mobile functionality without `ocsigen-start`, but it will be hard to maintain an Eliom version, so let's remove it.

PR #377, opened 2016-09-21
Add some comments in Makefile.options to avoid an error with ocsipersist

dannywillems:
If `sqlite3` (or `dbm`) is not installed, we have the error:
```
ocsigenserver: ocsigen:main: Fatal - Findlib package ocsigenserver.ext.ocsipersist-sqlite not found: maybe you forgot <findlib path="..."/>?
Makefile:58: recipe for target 'test.byte' failed
```

I put it in comments and not as OPAM dependencies to avoid to have too many OPAM dependencies and maybe the dev will always use `dbm` and not `sqlite`.

### DIFF
--- a/pkg/distillery/basic.camlp4/Makefile.options
+++ b/pkg/distillery/basic.camlp4/Makefile.options
@@ -28,6 +28,9 @@ CLIENT_PACKAGES :=
 LOCAL_STATIC = static
 
 # The backend for persistent data. Can be dbm or sqlite.
+# Be sure you installed the following OPAM packages:
+# - *dbm* if you use dbm --> opam install dbm.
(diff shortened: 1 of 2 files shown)


pkg/distillery/basic.camlp4/Makefile.options
```diff
@@ -28,6 +28,9 @@ CLIENT_PACKAGES :=
 LOCAL_STATIC = static
 
 # The backend for persistent data. Can be dbm or sqlite.
+# Be sure you installed the following OPAM packages:
```
vasilisp: Maybe "make sure you have the following packages installed"?

dannywillems:
@vasilisp Fixed.

vasilisp:
Thanks!

PR #378, opened 2016-09-21
Add comment to install ocsipersist in Makefile.options.

dannywillems:
@vasilisp Merged too fast :).

I hesitated to add it as OPAM dependency.

### DIFF
--- a/pkg/distillery/basic.camlp4/Makefile.options
+++ b/pkg/distillery/basic.camlp4/Makefile.options
@@ -29,6 +29,7 @@ LOCAL_STATIC = static
 
 # The backend for persistent data. Can be dbm or sqlite.
 # Make sure you have the following packages installed:
+# - ocsipersist --> opam install ocsipersist
 # - *dbm* if you use dbm --> opam install dbm.
 # - *sqlite3* if you use sqlite --> opam install sqlite3.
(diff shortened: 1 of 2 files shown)


vasilisp:
Not really :). We are still using the `ocsipersist` distributed with `ocsigenserver`. The separate package is not in OPAM, and does not even include the recently-developed Posgres backend.

dannywillems:
@balat asked me to add `ocsipersist` in the OPAM repository. It's plan to switch to the `ocsipersist` not distributed with `ocsigenserver`

vasilisp:
@dannywillems I am glad that you are doing this. But before we update the documentation, we need to put all the latest code in the Ocsipersist repo, update the build systems everywhere needed, update all OPAM files, and make sure everything plays nicely together.

dannywillems:
@vasilisp (I edited my last comment. The last sentence was a question, not an affirmation.).
Hum, ok. So do I wait to add it?

vasilisp:
@dannywillems yes, it is planned. Somebody has to step up and prioritize this task, and of course that somebody could be you if you are motivated :).

The comment would be misleading without a properly-released `ocsipersist`, so I would say wait.

dannywillems:
@vasilisp Ahha, it's not about motivation, but about time :D.

PR #379, opened 2016-09-22
Fix forms on client apps

vouillon:
We are serving from a local file. So, if the form target were over
https, we were not performing the local action but performing a post
request, which does not make sense with client apps.


vasilisp:
Looks good.

PR #383, opened 2016-09-28
Client-side URL and action fixes

vasilisp:
This PR fixes a few issues with client-side actions and URL management.

Notably,

* `Eliom_service.reload_action` and `Eliom_service.reload_action_https` drop the NA parameters
* after performing an action, the URL contains the NA parameters of the action, even though we have produced the page corresponding to the URL without these. This is for compatibility with the server semantics.

There is some ugliness related to URLs; unfortunately we work with a mix of relative and complete URLs, and JSOO `Url` doesn't like the relative ones, while `Eliom_lib.Url` is not all that complete. We need to work on a coherent URL library.


Drup:
> We need to work on a coherent URL library.

Like using [ocaml-uri](https://github.com/mirage/ocaml-uri) everywhere ? :)

PR #386, opened 2016-09-30
Fix empty path after Eliom_client.change_page

vasilisp:
This fixes a negligence  in #383 ; the `set_target [] l` in `change_page` was never meant to be committed, just to allow me to focus on something else at that point in time.


balat:
Thanks! I tested it. It works witg OS.

Issue #362, opened 2016-08-18
crashing OS X Safari's engine


pwbs:
Following discussion started at https://github.com/ocsigen/eliom/commit/7c199de220719fe9750d21502948e649cd5995e1

Summary: somewhere between https://github.com/ocsigen/eliom/commit/a1889d4696f3cf46a701b5f77d95b166d97cedeb  and https://github.com/ocsigen/eliom/commit/9ab810f4ab0e38968a5ca539f48d6e66d2e03e1b something (probably in the generated JS) broke Safari's engine (which triggers a `SIGSEGV`).

> Maybe the mere presence of lots of EBA code causes trouble. Does the `Eba_page.Make(Page_config)` in bar_page.eliom make a difference?

I'm having troubles understanding this question...

vasilisp:
> I'm having troubles understanding this question... 

If you remove the functor invocation `Eba_page.Make(Page_config)` from the `bar_page.eliom` in the `bar.zip` you attached, does Safari still segfault?

pwbs:
If I do that, it doesn't compile... `Error: Unbound module Bar_page.Opt`

pwbs:
I'm now back to trying to solve this issue. Replacing the functor invocation by
```ocaml
(* include Os_page.Make(Page_config) *)

module Opt = struct
  let connected_page _ () () =  Lwt.return
      (html (head (title (pcdata "")) []) (body [pcdata "plop"]))

end
```
still makes Safari crash.

PR #387, opened 2016-09-30
Make the PPX emit less sections

Drup:
Apparently, the eliom ppx is a bit too eager to put sections everywhere.
Those two small optimisations are backported from the eliom compiler. This is basically untested.

@vouillon, can you take a look?

### DIFF
--- a/src/ppx/ppx_eliom_client.ml
+++ b/src/ppx/ppx_eliom_client.ml
@@ -126,6 +126,12 @@ module Pass = struct
           [%e AC.str @@ file_hash loc]
     ][@metaloc loc]
 
+  let may_close_server_section item =
+    if Cannot_have_fragment.structure_item item
+    then []
+    else [close_server_section item.pstr_loc]
+
+
   let open_client_section loc =
     [%stri
       let () =
@@ -133,39 +139,30 @@ module Pass = struct
           [%e AC.str @@ file_hash loc]
     ][@metaloc loc]
 
+  let may_open_client_section loc =
+    if flush_injection ()
+    then [ open_client_section loc ]
+    else []
+
   (** Syntax extension *)
 
   let client_str item =
-    if not (must_have_section item && flush_injection ())
-    then [item]
-    else begin
-      let loc = item.pstr_loc in
-      [ open_client_section loc ;
-        item ;
-      ]
-    end
+    let loc = item.pstr_loc in
+    may_open_client_section loc @
+    [ item ]
 
   let server_str item =
-    if not @@ must_have_section item then []
-    else begin
-      let loc = item.pstr_loc in
-      register_client_closures (flush_client_value_datas ()) @
-      [ close_server_section loc ]
-    end
+    register_client_closures (flush_client_value_datas ()) @
+    may_close_server_section item
 
   let shared_str item =
-    if not @@ must_have_section item then [item]
-    else begin
-      let loc = item.pstr_loc in
-      let client_expr_data = flush_client_value_datas () in
-      (if flush_injection () then [open_client_section loc] else []) @
-      register_client_closures client_expr_data @
-      define_client_functions loc client_expr_data @
-      [ item ;
-        close_server_section loc ;
-      ]
-    end
-
+    let loc = item.pstr_loc in
+    let client_expr_data = flush_client_value_datas () in
+    may_open_client_section loc @
+    register_client_closures client_expr_data @
+    define_client_functions loc client_expr_data @
+    [ item ] @
+    may_close_server_section item
 
   let fragment ?typ:_ ~context ~num ~id expr =
 
--- a/src/ppx/ppx_eliom_client.ml
+++ b/src/ppx/ppx_eliom_client.ml
@@ -128,26 +128,34 @@ module Pass = struct
   (** Syntax extension *)
 
   let client_str item =
-    let loc = item.pstr_loc in
-    [ open_client_section loc ;
-      item ;
-    ]
+    if not @@ must_have_section item then [item]
+    else begin
+      let loc = item.pstr_loc in
+      [ open_client_section loc ;
+        item ;
+      ]
+    end
 
   let server_str item =
-    let loc = item.pstr_loc in
-    register_client_closures (flush_client_value_datas ()) @
-    [ close_server_section loc ]
+    if not @@ must_have_section item then []
+    else begin
+      let loc = item.pstr_loc in
+      register_client_closures (flush_client_value_datas ()) @
+      [ close_server_section loc ]
+    end
 
   let shared_str item =
-    let loc = item.pstr_loc in
-    let client_expr_data = flush_client_value_datas () in
-    open_client_section loc ::
-    register_client_closures client_expr_data @
-    define_client_functions loc client_expr_data @
-    [ item ;
-      close_server_section loc ;
-    ]
-
+    if not @@ must_have_section item then [item]
+    else begin
+      let loc = item.pstr_loc in
+      let client_expr_data = flush_client_value_datas () in
+      open_client_section loc ::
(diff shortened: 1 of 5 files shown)


vasilisp:
@pwbs could you test if this patch magically resolves #362 ? A bit far-fetched, I know, but the trouble somehow magically arose by the patterns produced by our PPX, and these patterns will now change.

Drup:
@vasilisp I don't think that's far fetched, since that's the reason @vouillon gave me to apply that optim. :)

vouillon:
Indeed, it seems to magically resolves #362 :), at least on the `os.pgocaml` distillery template.

vasilisp:
Great news!

Tthe code looks good.

Drup:
@vouillon I assume you tested on the besport website ? I would prefer some serious testing before we merge that :)

src/ppx/ppx_eliom_client.ml
```diff
@@ -152,9 +163,10 @@ module Pass = struct
       open_client_section loc ::
       register_client_closures client_expr_data @
       define_client_functions loc client_expr_data @
-      [ item ;
-        close_server_section loc ;
-      ]
+      [ item ] @
+      (if flush_injection () then
```
vouillon: The `open_client_section` should be removed. Not the `close_server_section`. Right?

Drup: Indeed! I'll fix that in the afternoon.

vouillon:
So, the second commit is broken.
And I could not reproduce the crash with the os.pgocaml distillery template, so I can't tell after all whether this fixes the problem...
I'm going to try with besport.

src/ppx/ppx_eliom_client.ml
Drup: Should now be fixed.

vouillon:
So, the crash still happens with the first commit only.
And the second commit is still broken (this can be reproduced with the os.pgocaml distillery template).

src/ppx/ppx_eliom_client.ml
```diff
@@ -128,7 +138,8 @@ module Pass = struct
   (** Syntax extension *)
 
   let client_str item =
-    if not @@ must_have_section item then [item]
+    if not @@ must_have_section item
+    || flush_injection () then [item]
```
vouillon: That should be `not (must_have_section item && flush_injection ())`

Drup: Oups. Fixed.

vouillon:
The besport website is still crashing on Safari with this change :(

But I think we are still emitting a lot more `close_server_section` than with the camlp4 syntax extension.

Drup:
The thing is, if you have this piece of code:

```ocaml
{client{
let x = 2
let y = 3 
}}
```
the camlp4 extension will consider that as only one section. The ppx consider it has two sections. The reason is to handle that kind of cases correctly:

```ocaml
{shared{
let x = 3 
let y = 3 + {client{ ~%x }}
}}
```
This doesn't work in the camlp4 extension, you must split in two different sections. In the ppx, it'll work directly, since sections are naturally split.

We can probably find a better scheme by analyzing things less locally. I'll try to think about it.

Drup:
Several new ideas:

1. If the declaration doesn't lead to any evaluation (in particular, is a function declaration), we can remove the call to `close`.
2. Successive server sections do not need to be separated by a `closer_server_section`. We can have one `close` after the last successive server section.
3. We emit many constant string that are equals. We could share them.

Drup:
I implemented idea 1 and 3.
Using ocsigen-start's `os.pgocaml` template as a benchmark. 

 |close_server_section|open_client_section|register_client_closure|Constant strings
---|---|---|---|---
Current|747|1053|217|6303
31bcbd8|665|963|217|6135
cd8e323|665|47|217|5219
6bd4f58|665|47|217|5219
4b12707|278|47|217|4832
7099dfb|278|47|217|4577

I can probably reduce it further by fine tuning `Cannot_have_fragment`. I'm a bit disappointed by the limited impact of idea 3.

Drup:
The javascript was compiled with ` -jsopt --noinline -jsopt --disable=shortvar -jsopt --pretty`.

Drup:
Further improvements on idea 1 to eliminate module elements. It looks quite good now. :)

 |close_server_section|open_client_section|register_client_closure|Constant strings
---|---|---|---|---
Current|747|1053|217|6303
288a58c|160|47|217|4546

Drup:
@vouillon Running `grep -o local/var/www/<app>/<app>.js -e '"[^"]*"'` on the `os.pgocaml` example is a bit striking, there is a lot of room for improvement. 

In particular, if we run `grep -o local/var/www/<app>/<app>.js -e '"[^"]*"' | sort | uniq -c | sort -nr` on the compressed version, we can see that there is *259* instances of `"number"`...

Drup:
(For the curious, the size impact is quite small, less than 5% without gziping)

vouillon:
The repetition of the string `"number"` is deliberate. More efficient code is produced for the expression `typeof x == "number"` than for `typeof x == y`  when `y` is bound to `"number"`.

vouillon:
The result is impressive. And Safari no longer crashes!

 |open_client_section|close_server_section
---|---|---
camlp4|3200|2627
ppx(master)|3328|2748
ppx_opt|374|484

There is still room for improvement. For file `os_services.eliom` in `ocsigen-start`, I think we only need one `close_server_section` and one `open_client_section`, but we get the following code:
```
close_server_section(eliom_compilation_unit_id_fhytq);
close_server_section(eliom_compilation_unit_id_fhytq);
close_server_section(eliom_compilation_unit_id_fhytq);
close_server_section(eliom_compilation_unit_id_fhytq);
close_server_section(eliom_compilation_unit_id_fhytq);
close_server_section(eliom_compilation_unit_id_fhytq);
close_server_section(eliom_compilation_unit_id_fhytq);
close_server_section(eliom_compilation_unit_id_fhytq);
close_server_section(eliom_compilation_unit_id_fhytq);
close_server_section(eliom_compilation_unit_id_fhytq);
close_server_section(eliom_compilation_unit_id_fhytq);
open_client_section(eliom_compilation_unit_id_fhytq);
var main_service$0=get_injection(_akR_,_akQ_,_akP_);
open_client_section(eliom_compilation_unit_id_fhytq);
get_injection(_akU_,_akT_,_akS_);
open_client_section(eliom_compilation_unit_id_fhytq);
get_injection(_akX_,_akW_,_akV_);
open_client_section(eliom_compilation_unit_id_fhytq);
get_injection(_ak0_,_akZ_,_akY_);
open_client_section(eliom_compilation_unit_id_fhytq);
get_injection(_ak3_,_ak2_,_ak1_);
open_client_section(eliom_compilation_unit_id_fhytq);
var connect_service=get_injection(_ak6_,_ak5_,_ak4_);
open_client_section(eliom_compilation_unit_id_fhytq);
var disconnect_service=get_injection(_ak9_,_ak8_,_ak7_);
open_client_section(eliom_compilation_unit_id_fhytq);
var action_link_service=get_injection(_ala_,_ak$_,_ak__);
open_client_section(eliom_compilation_unit_id_fhytq);
var set_password_service=get_injection(_ald_,_alc_,_alb_);
open_client_section(eliom_compilation_unit_id_fhytq);
get_injection(_alg_,_alf_,_ale_);
```

Drup:
@vouillon Nice! You can merge after you reviewed/when you feel like it, so that besport can avoid all those forks. :)

I'll do a separate PR if I get more improvements (with idea 2, in particular).

Drup:
@vouillon I would be interested by a comparison of the size of request data for besport too, if it's not too complicated to do.

vasilisp:
Well done @Drup :).

PR #388, opened 2016-09-30
Add break line to prevent right overflow

dannywillems:
See #330.


vasilisp:
Can you indent it properly while you are at it? `~service` is an argument to `make_uri`, and thus putting it in the same level as `~src` is confusing. Use `let` bindings if that helps.

dannywillems:
@vasilisp Oh yes! Fixed.

vasilisp:
Thanks!

PR #392, opened 2016-10-01
Eliom_client: the 'replace' argument was ignored at some places

vouillon:
Also, turn it into a mandatory argument, so that we no longer forget it.


vasilisp:
Looks good.

Issue #390, opened 2016-09-30
Redirections and reload_action


balat:
Hi,
There is still a problem with client side reload action.
Reload_action is supposed to remove na-parameters.
But if a na-service (implemented client-side) does a redirection like this:
```
      Eliom_registration.
        (appl_self_redirect
           Redirection.send
           (Redirection Eliom_service.reload_action))
```
This will create a loop ...

balat:
This redirection is still present in Ocsigen-start, file PROJECT_NAME_handler.eliom.
The temporary workaround is to replace the redirection by an action.

vasilisp:
@balat, I had missed this. (Notifications by e-mail don't work the way they used to!)

I will look into it.

vasilisp:
@balat I think the `no-srv-nesting` branch fixes the issue, but it has not yet received adequate testing. That branch will also allow us to simplify the `Eliom_client` code a bit.

balat:
Ok thanks. I'll test this.

balat:
I tested this and I still have a redirection loop, when opening a link in the app, when the mobile app is not launched.
I click on a link which opens in the app. The phone asks whether I want to open the link in the app or browser. I choose app. The app starts. And it asks again whether I want to open the link in the app.
As if it wanted to redirect to an external link ...

Still testing ...

balat:
Could it be that the redirection is producing an absolute URL, which is interpreted by Eliom as an external link?
Relative redirections have been allowed in HTTP only for a few years (2014).

How do client-side redirection work?

vasilisp:
The `send` function initiates an `Eliom_client.change_page` to the service we are redirecting to. Given that we know the concrete service, we call its client function without going through routing. So URLs matter only for the final `change_url_string` once we are done, and we only deal with the path and parameters.

vasilisp:
In my experiments, your code works correctly (unlike before). But I just call the redirecting NA service without routing, as would happen with a link. I will try to reproduce something similar.

vasilisp:
@balat would it be hard to test without the last commit, i.e., pin 02348807b33bf65020c1b979d85846cbc884e448 ? If that works I understand the problem.

balat:
ok I do that

balat:
Indeed it work without the last patch! ...

vasilisp:
We are almost there then :). I am fixing the last commit, without which there will be some noise in the history.

balat:
[...]

PR #395, opened 2016-10-05
No nested `change_page` calls for client services

vasilisp:
This PR simplifies the client service logic a bit.

Previously, to implement actions and redirections, we used to perform "nested" calls to `Eliom_client.change_page` and friends. This led to execution paths that were hard to get right. We now register the follow-up page change for later.

Fixes (I hope) #390 , among other issues.


balat:
I commented in #390

PR #397, opened 2016-10-06
Fix recent client value issues with the ppx

Drup:
In particular, fix https://github.com/ocsigen/ocsigen-start/issues/202
Thanks @vasilisp for noticing.

Wow, that was quite hard to debug. o_°

The predicate to remove `close_client_section` was inconsistent between client and server code generation on [this specific code pattern](https://github.com/ocsigen/ocsigen-start/blob/d60b89105786608f101fdf4d64cce28c7dce98e9/template.distillery/PROJECT_NAME_otdemo.eliom#L40-L74), due to the order of AST transformation.

We now compute the predicate before any further transformation of the AST, which should prevent similar issues. The predicate for client section should hopefully not suffer similar issues (since it's computed by the traversal itself, on the fly).

### DIFF
--- a/src/ppx/ppx_eliom_client.ml
+++ b/src/ppx/ppx_eliom_client.ml
@@ -96,6 +96,9 @@ module Pass = struct
     | [] -> []
     | _ -> [Str.eval (AC.sequence registrations)]
 
+  (* We hoist the body of client fragments to enforce the correct scoping:
+     Identifiers declared earlier in the client section should not be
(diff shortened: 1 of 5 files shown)


Drup:
Just to check all of this was not for nothing:

 |close_server_section|open_client_section|Constant strings
---|---|---|---|---
Current|747|1053|6303
after first PR|160|47|4546
after second PR|157|48|4542

(Pretty sure  `ocsigen-start` changed a bit in the middle, but we can see it's similar. :)

PR #398, opened 2016-10-06
Make Eliom_parameter user types injectable

vasilisp:
This patch allows us to inject user parameters, and by extension services that use them.

A prerequisite for removing client-side service creation (for which the patch is ready).

A bit hackish in places.

### DIFF
--- a/src/lib/eliom_common.server.ml
+++ b/src/lib/eliom_common.server.ml
@@ -1365,3 +1365,38 @@ let get_secure secure sitedata =
   match secure with
   | None -> sitedata.secure_cookies
   | Some s -> s
+
+module To_and_of_shared = struct
+
+  (* FIXME : work-around for weak polymorphism in create :( *)
+  type wrapper
+
+  type 'a t = {
+    server  : 'a to_and_of ;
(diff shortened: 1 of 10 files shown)


vasilisp:
We do not use shared values. That would complicate the life of existing server-only users, and it would complicate our internal module dependency dependency graph as well.

src/lib/eliom_common.server.ml
```diff
@@ -1365,3 +1365,36 @@ let get_secure secure sitedata =
...
+    wrapper : wrapper
+  }
+
+  let wrapper : wrapper = Obj.magic @@
+    Eliom_wrap.create_wrapper @@ function
+    | {client = Some tao} ->
+      tao
+    | {client = None} ->
+      failwith
+        "Cannot wrap to_and_from (missing client values for user_type?)"
```
Drup: I think you should find a way to make that error more precise. For example, it should probably give the name of the eliom parameter.

Drup:
Nice, that was an annoying issue with this API. I only took a glance at the code, it looks ok.

PR #400, opened 2016-10-10
Remove client-side server creation

vasilisp:
This PR removes client-side service creation. Services can still be used on the client, but they have to be injected (`~%service`).

Creating a service on the client is not necessarily meaningful, because we cannot possibly now if there exists a compatible server-side service listening on the right path for the right parameters.

`ocsigen-start`, `ocsigen-toolkit`, and `ocsigen-widgets` are all ready for this change. @balat @vouillon you may need to update BeSport :).


vasilisp:
@balat @vouillon can you check whether your codebase is compatible with this PR?

PR #404, opened 2016-10-19
Replaces tabs with spaces and remove some empty lines

dannywillems:
I suppose spaces it's the convention. It's so horrible to see different coding conventions when you try to understand in deep how Eliom works.

The same work must be done in other ocsigen projects. I saw tabs in some ml files in jsoo.

I didn't check for empty lines.

### DIFF
--- a/src/lib/client/eliom_client.js
+++ b/src/lib/client/eliom_client.js
@@ -72,7 +72,7 @@ var caml_unwrap_value_from_string = function (){
           var size = (code >> 4) & 0x7;
           var v = [tag];
           if (size == 0) return v;
-	  intern_obj_table[obj_counter] = v;
+          intern_obj_table[obj_counter] = v;
           stack.push(obj_counter++, size);
           return v;
         } else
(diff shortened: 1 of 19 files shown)


Drup:
I'm surprised, didn't we normalized with ocp-indent not-so-long ago ?

dannywillems:
Some tabs was there since several years (2012-2013).

vasilisp:
@dannywillems, I know that tabs are annoying, but fixing them all at once would mess up `git blame`.
`git blame` is invaluable; looking at the associated commits has helped me quite a few times understand why things are the way they are :).

I personally re-indent when making major changes to a module, but just accept the previous possibly-ugly style otherwise.

Drup:
@vasilisp `git log -L linestart,lineend:foo.ml` ? You get the whole history that way, which is much better anyway ...

dannywillems:
I agree that `git blame` is invaluable and since I know it, I use it a lot. The problem with doing the style migration only when there is major changes is not a good solution because maybe it will never be done.

Both solutions are ok for me.

PR #412, opened 2016-10-31
Remove extensions (Eliom_{atom,openid})

vasilisp:
ocsigen/ocsigen-start#270 reminded me that we have some OpenID code in Eliom. We also have some stuff about Atom feeds, and a small module for doing server-to-server requests.

Does anyone remember about this stuff? I suggest we should remove them. They frequently get in the way when updating the APIs, and they can easily be used outside the Eliom tree if one is so motivated.


dannywillems:
@vasilisp Are you sure nobody use these modules?

dannywillems:
I think server-to-server communications are useful. A high-level interface for fetching data with `Ocsigen_http_client`is interesting (I use it in https://github.com/ocsigen/ocsigen-start/pull/270).

vasilisp:
@dannywillems I am not saying that Atom, OpenID, and server-to-server are not useful. They are just out of context for Eliom and they can be easily maintained externally if anybody cares.

PR #413, opened 2016-11-01
Support getting templates from multiple directories, listed in $ELIOM…

FlorentBecker:
…_DISTILLERY_PATH

This allows the installation of other templates for eliom-distillery on systems where eliom's dir is not writeable (for instance, because it has been installed through system packages).

### DIFF
--- a/src/tools/distillery.ml
+++ b/src/tools/distillery.ml
@@ -6,7 +6,11 @@ let reserve_project_name_filename = ".eliomreserve"
 (* File containing files which must be ignored while copying a template. *)
 let eliomignore_filename          = ".eliomignore"
 
-let distillery_basic              = "basic.ppx"
+let eliom_template_dir = Findlib.package_directory "eliom.templates"
+
+let distillery_basic = ("basic.ppx", eliom_template_dir)
+
+let template_path (tname, tpath) = tpath ^ "/" ^ tname
 
 (* Returns all lines of [file] as a string list. Returns an empty list if [file]
  * doesn't exist.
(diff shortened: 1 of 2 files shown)


Drup:
Very nice! Could you specify the format of `ELIOM_DISTILLERY_PATH` (`:` separated list of absolute paths) in the documentation ?

FlorentBecker:
done.

dannywillems:
@FlorentBecker
I think it's not a bad idea to use this environment variable. Maybe it can be mixed with the idea of having a command line argument to add and remove a template with `eliom-distillery`.
This argument could accept paths but also URL.

I implemented the same feature [here](https://github.com/dannywillems/eliom-distillery/blob/master/distillery.ml) but not with the path.

FlorentBecker:
@dannywillems I think your feature and mine are complementary: yours is about using eliom-distillery to install more templates, mine is for having the distillery recognize externally installed templates. Both have their use.

A third solution would be to use findlib, and consider any package named *.distillery-template to be a template. That way, templates can be installed and found using existing tools.

src/tools/distillery.ml
```diff
@@ -207,19 +211,29 @@ let preds () = [
...
       if f = ".." || f = "."
-      then aux rl
-      else aux (f::rl)
+      then aux rl (path, dir)
+      else
+        aux ((f, path)::rl) (path, dir)
     with
       | End_of_file -> Unix.closedir dir; rl
-  in aux []
+  in List.concat (List.map (aux []) dirs)
```
vasilisp: I suspect all this would be cleaner with a `fold_left` over `get_templatedirs ()` (with the `opendir` and concatenation happening inside), but no big deal.

vasilisp:
Thanks @FlorentBecker! Looks good for now, merging.

@FlorentBecker @dannywillems, feel free to propose improvements with follow-up PRs.

PR #414, opened 2016-11-02
Distillery: remove warnings deprecated

dannywillems:
There are also others warnings about deprecated functions in the library.


vasilisp:
@dannywillems, I am aware of these warnings, but `lowercase_ascii` only exists since `4.03.0`. The cleanest solution is to live with the warnings for now, I am afraid

dannywillems:
@vasilisp Oh, I didn't know it was since `4.03.0`... I agree with you.
I close.

PR #415, opened 2016-11-02
Add Eliom_lib.Url.{path_of_url,path_of_url_string}

vasilisp:
Minor URL-related cleanup.

I added `Eliom_lib.path_of_url_string` by basically de-inlining variants of it from different modules.

The intermediate `Eliom_lib.path_of_url` function which operates on JSOO `url` types looks useful, so it is also exported.

### DIFF
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -772,47 +772,12 @@ let route ~replace ?(keep_url = false)
     Eliom_request_info.get_sess_info := r;
     Lwt.fail e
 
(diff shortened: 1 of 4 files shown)


dannywillems:
@vasilisp Is it related to https://github.com/ocsigen/ocsigenserver/pull/113?

vasilisp:
@dannywillems I forgot about ocsigen/ocsigenserver#113 :). Related as both extract the path. Different in that this one is tied to the JSOO `Url` module and it needs to handle relative URLs.

PR #418, opened 2016-11-09
eliom-distillery: permit copying files without modifications

### LINKED ISSUE #417: Eliom distillery corrupts binary files (URGENT for Eliom 6)
It adds a newline at the end (and could possibly do other changes).
This breaks the fonts in Ocsigen Start, for the mobile app.

A solution could be to use a special suffix for binary files and just copy them without change.

vasilisp:
`eliom-distillery` now permits copying files (those with the `.eliomverbatim` suffix) without modifications.

Fixes #417.

### DIFF
--- a/src/tools/distillery.ml
+++ b/src/tools/distillery.ml
@@ -91,6 +91,23 @@ let mkdir_p path_str =
   in
   aux [] (split_path path_str)
 
+let copy_file_plain input_name output_name =
+  mkdir_p (Filename.dirname output_name);
+  let buffer_size = 8192 in
+  let fd_in = Unix.openfile input_name [O_RDONLY] 0
+  and fd_out =
+    let {Unix.st_perm} = Unix.stat input_name in
+    Unix.openfile output_name [O_WRONLY; O_CREAT; O_TRUNC] st_perm
+  and buffer = String.create buffer_size in
+  let rec copy_loop () =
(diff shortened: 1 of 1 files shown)


balat:
Thanks! I'm testing this with Ocsigen Start and I will merge if it works.

vasilisp:
It is broken, wait!

vasilisp:
@balat it should be fixed now.

balat:
I'm compiling (but it takes time, I had to create a new opam switch).
Can you add some lines about this in Eliom's reference manual please?
Actually, I'm not sure whether eliom-distillery is documented for template authors (?)

vasilisp:
There is no documentation for template authors. But should we really encourage external authors to provide templates? I am not really convinced that it's a good idea to generalize their use :).

balat:
It works!

Why do you think it's not a good idea?

vasilisp:
@balat, templates are OK the way we are using them now (co-developed with Eliom and OS). But generalizing their use would amount to a half-baked distribution mechanism, and there are plenty of these already. In addition, templates pose maintainability problems. It is hard to update the templates when Eliom changes, and even harder to update template-derived apps when the template changes.

I would much prefer it if anything worth sharing became available as an OPAM/ocamlfind package with a friendly API :).

PR #421, opened 2016-11-14
Fix #419

### LINKED ISSUE #419: change page to reload action broken when there is no client side implementation of the service
```
let%client reload () =
  Eliom_client.change_page ~service:Eliom_service.reload_action_hidden () ()
```
produces a 404 error when the current page is implemented server-side only.

vasilisp:
Fixes #419. CC @balat.

It is trickier than I thought. If the current service is not registered client-side (in which case our routing-based reload fails), we want to do what we would normally do for a server-side service (e.g., stay in the app if possible). I didn't want to replicate the logic in `Eliom_client`, so I added a flag `~ignore_client_fun` to `Eliom_client.change_page`. This permits us to resort to an
```ocaml
Eliom_client.change_page ~ignore_client_fun:true ~service:Eliom_service.reload_action () ()
```
as a server-based "fallback" reload.


balat:
If I understand well, with `~ignore_client_fun:true` the client service will be ignored in any case?

I think this is a problem for reload_action, because we usually don't know whether there is a client side implementation or not ...

Do I miss something?

vasilisp:
The `~ignore_client_fun` part is only used as a last resort, when client routing has failed.

Before we get there, we try to do client routing to the current URL (with or without the NA parameters, as appropriate). Assuming that there is a client-side implementation, client routing succeeds and we are done.

If client routing fails, I tried to replicate what would happen with reload actions before we implemented client functions for them, and the `~ignore_client_fun` is a way to do it.

vasilisp:
The latest version uses a new `Eliom_client.server_reload_action` for the server-based reload. `~ignore_client_fun` has been removed.

PR #422, opened 2016-11-15
Secure cookies: doc + code cleaning

balat:


src/lib/eliom_common.server.mli
```diff
@@ -711,7 +711,7 @@ type eliom_js_page_data = {
 
 val get_site_dir : sitedata -> Url.path
 val get_site_dir_string : sitedata -> string
-val get_secure : bool option -> sitedata -> bool
+val get_secure : secure_o:bool option -> sitedata:sitedata -> unit -> bool
```
vasilisp: To be honest, I don't think that `_o` improves clarity, because we can already see that it is an option from the type. Omitting it would just make passing around `secure` more concise.

src/lib/eliom_state.server.ml
```diff
@@ -94,7 +97,7 @@ let set_global_service_state_timeout
   let sitedata =
     Eliom_request_info.find_sitedata "set_global_service_timeout"
   in
-  let secure = Eliom_common.get_secure secure sitedata in
+  let secure = Eliom_common.get_secure secure sitedata () in
```
vasilisp: We unfortunately have warning 6 disabled. It would be good to label `secure`.

src/lib/eliom_common.server.mli
balat: This is an internal function, and naming it like that helped a lot to check bugs in that code
(because we have optional `secure` sometimes, sometimes `secure` mandatory and bool, sometimes `secure` mandatory and bool option...) 
Anyway this commit is not really important. I did it while tracking a difficult bug and find that it will help in the future (because the code for secure cookies is somewhat tricky).

src/lib/eliom_state.server.ml
balat: Yes. We can remove this.
Actually may be we don't want that commit at all.
The important one is the other one, because it points out a very delicate configuration problem ....

vasilisp: @balat, no need to remove the commit. If my comment was not clear, I was suggesting `Eliom_common.get_secure ~secure_o:secure` (the compiler accepts the unlabeled version, and we have disabled the warning).

vasilisp:
Commit the version you prefer, my comment was minor and purely aesthetic :).

balat:
I think it's better now. I kept the ~secure_o because I was lost with all these secure stuff.
I added the labels.

The most problematic discussion is about the other commit.
We had `<securecookies value="true"/>` at a wrong place in the configuration file and Os_tips was broken ... I considered removing the possibility to have this option in `<host>` section (and keep this option in `<extension ... "eliom">`), but this would make it impossible to have different settings for different sites.
I hope the warning in the documentation is enough ... This produces bugs very difficult to fix :-/

PR #424, opened 2016-11-16
Support PPX in eliompp

### LINKED ISSUE #423: eliompp doesn't work with PPX
The ocsigen-start documentation doesn't show the contents of client-side modules. I think it boils down to `eliompp` (called by `eliomdoc`), which does not handle PPX at all. It finds no `{section{ }}` annotations, so it assumes everything is server code.

That's critical. We can't release `ocsigen-{toolkit,start}` with the documentation in that state.

For the documentation of Eliom itself, we don't use `eliomdoc` at all (but `ocamldoc` with the appropriate PPX). This has its own issues: comments that are for the wrong side, e.g., the last line in https://ocsigen.org/eliom/dev/api/client/Eliom_cscache . That needs to be fixed also.

vasilisp:
It is not pretty, but I think it works. Already [tested with Ocsigen Start](https://github.com/ocsigen/ocsigen-start/commit/da23a49a145410e9a3a9ff584c7ed9e27da43891).

Fixes #423.

### DIFF
--- a/src/tools/eliompp.ml
+++ b/src/tools/eliompp.ml
@@ -37,9 +37,27 @@ let pretty_print_header filename =
   print 1 "<command-line>";
   print 1 filename
 
+let current_section () = !Eliompp_lexer.section_idt
+
+let we_are_ppx () = !Eliompp_lexer.we_are_ppx
+
+let for_server () =
+  match current_section () with
+  | `Server | `Shared -> true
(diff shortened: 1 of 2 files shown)


vasilisp:
Merging to get the infrastructure going. Let me know if anything looks weird in the docs.

PR #427, opened 2016-11-22
Doc: chapter mobile app + scalability

balat:
This is a draft. I'd appreciate a careful proofreading.


doc/manual-wiki/mobile-apps.wiki
```diff
@@ -0,0 +1,73 @@
...
+The best way to start with this feature is to install
+<<a_manual project="ocsigen-start"|Ocsigen Start>>.
+It contains all the instructions and a template with code samples
+that you can adapt to your needs.
+
+Here are the main principle for writing multi-platform apps with Eliom:
+
+* Use responsive Web design to make your app work with any screen size.
+* Register all your services in shared sections. The mobile app will
+use the client side implementation. The Web app will generate the pages
```
vasilisp: _client-side_ (missing dash)

doc/manual-wiki/mobile-apps.wiki
```diff
@@ -0,0 +1,73 @@
...
+server-side during the first request, and by default, subsequent pages will
+be generated client-side.
+* Be careful with the order of evaluation, which is different when pages
+are generated client-side. Client sections sent by the server will be executed
+when the page is ready and already diplayed. But if the page is generated
+client-side, client-section will be executed straight-away. You may sometimes
```
vasilisp: _client sections_

doc/manual-wiki/mobile-apps.wiki
```diff
@@ -0,0 +1,73 @@
...
+want to delay their execution to when the element is inserted in the page
```
vasilisp: _to when_ -> _until_

vasilisp:
There is also a sparse tutorial page: http://ocsigen.org/tuto/dev/manual/mobile. In its current version it only says "go use Ocsigen Start" (in more words). Should I remove it?

balat:
Thank you for your comments!

We need a chapter called "Mobile apps" because that is something people will look for. And they won't understand "client-side services".

I would not merge the two, because it is possible to use client-side services even if you are not doing a mobile app. It makes it possible to do page generation from client side.

We'll think about the tutorial when the manual is complete.

balat:
About your first comment (about speaking about OS in Eliom's documentation):
Let's write what is useful for user ;)
It would be great to do a detailed explanation of mobile apps with Eliom. But it is very difficult o write (we are not ready for that). I think that the right way (right now) to write a mobile Eliom app without Ocsigen Start, is to install Ocsigen Start and try to mimic what it does ...

vasilisp:
`Eliom_notif` is out-of-sync with `Os_notif` and it doesn't support multiple servers. I was thinking of completely removing it for now and fixing it later (because picking up new tasks delays the release).

balat:
Jan is currently looking at Eliom_notif (trying to update it with his changes).

balat:
Careful proof reading is also needed for the introduction.

vasilisp:
I read the intro, I committed a couple of minor changes.

balat:
I'm proof reading chapter services.
I have a problem with section "Unregistrable services".
It links to itself, and the following sections should probably be sub-sections?

vasilisp:
Right, I moved text around leading to a now-redundant link. I will fix it.

vasilisp:
@balat while you are at it, are you happy with describing what was previously a non-attached coservice as "pathless", and by using attached only for attached coservices? These are the more prevalent changes in that page.

balat:
Yes I like the new description of services. It's much simpler than before. :)

balat:
The manual seems now complete to me now.
I hope we did not forget some module in the index (manual and API).
I added some of them (Eliom_notif, cscache...)

@jrochel can you have a look at chapter scalability (in this PR), and also section "notifications" in chapter clientserver-communication? Thanks!

jrochel:
I don't see any problems. Maybe you can just clarify that `scalability' here means running one application on multiple servers.

vasilisp:
We can merge this when you want. We may need to do adjustments after we see the output online.

vasilisp:
@balat merging, we can always improve it later.

PR #431, opened 2016-11-27
Fix server-client type checking for PPX under ocamlbuild.

paurkedal:


Drup:
I'm surprised that doesn't work already. Don't we use that plugin to build eliom itself ?

paurkedal:
Yes, I'm digging into you build setup wondering why it works.  And as you can see Jenkins spotted my mistake, which was shadowed by an installed version of eliom.  I've being using this patched version myself a few weeks after discovering that server-client type-checking didn't work. Apparently it works in the eliom source dir, just tried to introduce a type error, and it was rejected.  Will make another attempt with the unpatched version.

paurkedal:
Seems type-checking isn't complete in the eliom source, as it still builds with the following tweak:
```
diff --git a/src/lib/eliom_form.eliom b/src/lib/eliom_form.eliom
index e9ec9d3..a62f413 100644
--- a/src/lib/eliom_form.eliom
+++ b/src/lib/eliom_form.eliom
@@ -671,3 +671,8 @@ module Make (Html : Html) = struct
       contents getparams
 
 end
+
+[%%server.start]
+let test = 10
+[%%client.start]
+let test : float = ~%test
```

vasilisp:
Thanks @paurkedal!

I guess the type checking hasn't been happening since the recent transition to PPX.

However, I don't yet understand the impact of the `side` argument. We only need the `type_mli` files when compiling client code, the checking is not bi-directional. Would the result be equivalent to not tagging the server file?

paurkedal:
@vasilisp Yes, you're probably right that we don't need to pass -type for the server side compilation, but we'd still need to use different options for server and client, right?

We also need to find a solution so that it also works inside the eliom source tree. My first though is to create a fake META to use for the build. But maybe we need to support non-ocamlfind build also, in any case?

vasilisp:
> @vasilisp Yes, you're probably right that we don't need to pass -type for the server side compilation, but we'd still need to use different options for server and client, right?

Yes, certainly.

> We also need to find a solution so that it also works inside the eliom source tree. My first though is to create a fake META to use for the build. But maybe we need to support non-ocamlfind build also, in any case?

Wouldn't it be simpler to pass the flags to the PPX executables directly? I think this can happen [around here](https://github.com/ocsigen/eliom/blob/6d520df5306d4ff10a396d57b83637c8234f59da/build/build.ml#L43), but I haven't tried yet. We could use a Boolean to distinguish between internal and external builds in the `Ocamlbuild_eliom` functors to work around the issue revealed by the Jenkins build.

This is release-critical for Eliom 6.0, so it needs to be resolved really soon :).

vasilisp:
> But maybe we need to support non-ocamlfind build also, in any case?

I would say let's not worry about that, it would be of very limited utility.

paurkedal:
I think the natural thing to do would be change the command line in build/build.ml to
```ocaml
S [A "-ppx"; Quote (S [P (path ^ name ^ "_ex.native"); T ppx_tags])]
```
and let `Ocamlbuild_eliom` pass the `-type` path for `T ppx_tags`.  However, this can't work since `ppx_tags` won't include the `file:` tag, so `flag_infer` won't be able to target precisely the command line which needs a particular type file.

The option I can see, is to let `Ocamlbuild_eliom` pass the full ppx command line for internal build.  Maybe adding a `val eliom_ppx_dir : string option` to `INTERNAL` indicate an internal build and where the extensions are. At this point `with_package` may start to get redundant, so I may change that as well. I'll make another PR this evening, just want to spill the idea in case you have other options.

vasilisp:
Thanks for investigating! You are right. I tried to modify `build/build.ml` and I realized myself that we can't obtain the right `type_mli` argument. As of now, I have no idea other than what you describe.

paurkedal:
I pushed a new verison, but now there is probably many type annotations to add to avoid existential types in the inferred interface before merging the PR.

About that, I think most of the existential types are due to the use of references in the generated code.  I haven't look into the details, but I suspect they are assigned once and then only read? If so, maybe it would be possible to have the generalised by e.g. using lazy instead? I can open a new issue if you think it may be worth looking into.

vasilisp:
So far we have been assuming that these are truly polymorphic. In the Camlp4 versions of the tools, we can parse `'_a` (Camlp4 accepts it in input files, unlike the OCaml parser), and we do internal transformations to get rid of the underscore. In the PPX-enabled `eliomc`, we achieve something similar with a [`sed` hack](https://github.com/ocsigen/eliom/blob/master/src/tools/eliomc.ml#L202).

vasilisp:
> About that, I think most of the existential types are due to the use of references in the generated code. I haven't look into the details, but I suspect they are assigned once and then only read? If so, maybe it would be possible to have the generalised by e.g. using lazy instead? I can open a new issue if you think it may be worth looking into.

This is used as an ad-hoc type inference mechanism. See [a mailing list message](https://sympa.inria.fr/sympa/arc/ocsigen/2016-06/msg00025.html) for an explanation. I do not immediately see what we could do with `Lazy`. If you have immediately-implementable ideas, that would be great. However, I am not sure there is much potential in improving our hacks; we should rather focus on a [true type checker](https://github.com/ocsigen/ocaml-eliom) :).

paurkedal:
Thanks for background. Haven't had a proper look at it yet, but for now adding the sed hack to the ocamlbuild plugin seems like the best option.

paurkedal:
I lifted the sed transformation from ocamlc, and low and behold, no hard type errors. There is a number of warnings about possible wrongly inferred types, but I guess that's to be expected at this stage.  Let me know if you want some adjustments before merging.

vasilisp:
@paurkedal much appreciated! I will review carefully later today.

vasilisp:
It inspires enough confidence to merge immediately, so let's do that and test for a couple of days before we release.

The type warnings look normal. Before this PR we were just hiding them under the carpet. But I will see if some of them can be resolved.

PR #439, opened 2016-12-08
Remove calendarlib dependency

vasilisp:
I just realized that the dependency on `calendarlib` is redundant, as witnessed by the Jenkins build of this PR, which will succeed.

Unfortunately, to build without `calendarlib` we need a tiny modification of `src/_tags`. That is, to fix the version in `opam-repository`, we would either need a bugfix release with a new tarball, or a patch in `opam-repository`. It is not too critical, so I think we should just aim for a bugfix release in a month or so (but do it this time).

### DIFF
--- a/opam
+++ b/opam
@@ -15,7 +15,6 @@ depends: [
   "ppx_tools" {>= "0.99.3"}
   "js_of_ocaml" {>= "2.8.2"}
   "tyxml" {>= "4.0.0"}
-  "calendar"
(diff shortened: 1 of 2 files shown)


vasilisp:
In fact, the PR build didn't fully succeed because His Jenkinsness [decided to not cooperate](https://ci.inria.fr/ocsigen/job/Eliom-PR/521/console) (`java.net.SocketTimeoutException`, that's new). But it did build Eliom successfully :).

PR #445, opened 2016-12-12
Basic Reason support

vasilisp:
This PR provides very rough and mostly untested Reason support.

I added a `-reason` flag to our compiler wrappers. It is implemented by calling `refmt` as a preprocessor.

There is a new `basic.reason` template that uses Reason syntax.


vasilisp:
There is a problem with type-checking client values (and probably injections). It is because the new preprocessing pipeline introduces unpredictable intermediate filenames, impacting our auto-generated names for client fragments. To be fixed tomorrow.

src/tools/utils.ml
```diff
@@ -362,6 +364,12 @@ let get_ppopts ~impl_intf file =
     type_opt impl_intf file @ !ppopt
 
 let preprocess_opt ?(ocaml = false) ?kind opts =
+  let refmt () =
+    if !use_refmt then
+      [ "-pp"    ; "refmt -parse re -print ml" ]
```
Drup: Should we try to provide a clean failure when reason is not installed ?

Drup:
Nice work!

vasilisp:
@Drup, does e706f1d0602f849256507030ea627430525f0fa5 look line a sane approach?

vasilisp:
See #504.

PR #446, opened 2016-12-12
Eliom_client: do not execute onload after ocaml services (server fuctions)

balat:
Please read carefully (only 6 lines).
@vouillon @vasilisp 

The onload was done after RPCs which is wrong.
I hope I removed the right lines (and that it will not break anything).
Do not release this in 6.0 branch before testing on alpha.besport.com...

### DIFF
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -404,11 +404,10 @@ let call_ocaml_service
       get_params post_params in
   let%lwt () = Lwt_mutex.lock load_mutex in
   set_loading_phase ();
-  push_onload ();
   let%lwt content, request_data = unwrap_caml_content content in
   do_request_data request_data;
   reset_request_nodes ();
-  let load_callbacks = flush_onload () @ [broadcast_load_end] in
+  let load_callbacks = [broadcast_load_end] in
   Lwt_mutex.unlock load_mutex;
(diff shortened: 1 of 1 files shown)


src/lib/eliom_client.client.ml
```diff
@@ -402,15 +402,9 @@ let call_ocaml_service
       ?keep_nl_params ?nl_params ?keep_get_na_params
       ?progress ?upload_progress ?override_mime_type
       get_params post_params in
-  let%lwt () = Lwt_mutex.lock load_mutex in
```
vasilisp: Removing this looks suspicious. @vouillon added the lock in 123bfc49 to "prevent an OCaml service to be called while loading a page".

vasilisp:
Not running the callbacks look good to me. Why would `onload` be meaningful for OCaml services? But I think we need the lock. @vouillon?

src/lib/eliom_client.client.ml
vasilisp: The ocaml service itself is not a page, but the lock could apply in scenarios that involve nesting of service calls. It sounds like we would somehow end up calling an OCaml service in the load phase of another page, and trouble would ensue?

balat:
I pushed a much more conservative version, discussed with Jérôme, which just removes onload.

PR #460, opened 2017-01-03
Notifs: cleaning Eliom_notif (first step)

balat:
@vasilisp I removed some weird things in the cade. Read each commit separately (and carefully). I think it does not change the semantics at all.

There are still things to be checked. 

1. For example the `clean` function that was supposed to be asynchronous by was not. (Why do we need that? do we need to call that in BS or OS?
2. In the original version `notif_e` as a triple, not a pair, and there was an explanation why. How is this solved now?

### DIFF
--- a/src/lib/eliom_notif.server.ml
+++ b/src/lib/eliom_notif.server.ml
@@ -23,7 +23,7 @@ module type S = sig
   end
   val notify : ?notfor:[`Me | `Id of identity] -> key -> server_notif -> unit
   val client_ev : unit -> (key * client_notif) Eliom_react.Down.t Lwt.t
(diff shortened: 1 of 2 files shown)


vasilisp:
> For example the clean function that was supposed to be asynchronous by was not. (Why do we need that? do we need to call that in BS or OS?

It was asynchronous because an initial version was doing the cleanup periodically. See discussion in #331.  If the redundant entries in the table are of no concern, there is no need to call cleanup.

I will look into the triple vs. pair question, I am catching up on various things :).

vasilisp:
I am quite late, but the commits look fine.

Issue #464, opened 2017-01-09
Documentation on compiling with Oasis needs an update


shepard8:
If I try to compile the first step from the [widget tutorial](https://ocsigen.org/tuto/manual/tutowidgets) using Oasis, I run into lots of undocumented problems. I try to summarize them here. I followed the steps of the [workflow-compilation](https://ocsigen.org/eliom/6.1/manual/workflow-compilation) manual page.

- First of all it is not clear which parts of the ocamlbuild section have to be followed, and which are not. I think it could be useful to repeat information so that those interested in using Oasis (I use Oasis because it usually makes things easier) do not have to guess what is the useful information.
- When running `oasis setup` with the _oasis example given, whence directions in the comments have been applied, warnings are issued:
> W: Cannot find source file matching module 'server/Main' in library server.
> W: Use InterfacePatterns or ImplementationPatterns to define this file with feature "source_patterns".

I think it should at least be mentioned so that users do not think they are doing something wrong.
- It is not said that the existing dispatcher has to be removed in `myocamlbuild.ml`.
- Running `ocaml setup.ml -build` also issues these warnings (twice).
- Even when adding
> <*.eliom>: eliom_ppx
> <*.eliomi>: eliom_ppx

as suggested by the manual when PPX is used (which is the case with the widget tutorial), the following error is issued when building (`ocaml setup.ml -build`):
> \+ /home/shepard/.opam/4.02.3/bin/ocamlfind ocamldep -package eliom.server -package eliom.syntax.predef -package eliom.syntax.type -syntax camlp4o -modules src/type/main.ml > src/type/main.ml.depends
> File "src/main.eliom", line 15, characters 0-1:
> Parse error: [semi] expected after [str_item] (in [implem])
> File "src/type/main.ml", line 1:
> Error: Error while running external preprocessor
[...]

PR #467, opened 2017-01-09
Improved documentation. See issue 464.

shepard8:
Documentation update proposal according to remarks given in #464.

### DIFF
--- a/doc/manual-wiki/workflow-compilation.wiki
+++ b/doc/manual-wiki/workflow-compilation.wiki
@@ -104,7 +104,6 @@ end)
 
 let () = Ocamlbuild_plugin.dispatch M.dispatcher
 >>
-(if you are using OASIS, see also the next section)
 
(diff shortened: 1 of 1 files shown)


vasilisp:
Thanks! Looks good, but I'd prefer to go through your instructions and make sure everything works. Be a bit patient :).

shepard8:
Hello! For sure, take your time. By the way I'm still unsatisfied with the way users have to configure OASIS. This seems to be a lot of complications while using command line this is (as you know) just two lines:

```
eliomc -a -o server.cma -ppx file1.eliom file2.eliom ...
js_of_eliom -o client.js -ppx -package lwt.ppx -package js_of_ocaml.ppx file1.eliom file2.eliom ...
```

I tried to use the custom plugin of oasis but it seems it has to be used globally, which I don't want. I'm still looking for an easier way to setup OASIS, including creating a new plugin for OASIS.

Issue #465, opened 2017-01-09
Wrong URL for pathless services (non-attached coservices) in mobile apps


balat:
A request on a pathless service is done with URL `http://<server>/<local_path>`,
where `local_path` is the local path of the app on the mobile device filesystem (for example for RPCs).

This is wrong because:

- We don't want to send this info to the server
- In the case where the Eliom site is in a subdirectory of the server, the service is not found.

Solution discussed today  (with @vasilisp, @vouillon):

Eliom_client.init_client_app should remember the local path and, when is_client_app, pathless service and absolute, Eliom_uri should replace the local path by the Web site root. This will probably require to add a JS variable __eliom_app_path.

PR #468, opened 2017-01-12
Use a given path for absolute service calls

vasilisp:
See #465 and ocsigen/ocsigen-start#400.

I am not sure which parts of `Eliom_uri` need to be modified. I have only modified the execution path that happens for server functions.

### DIFF
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -976,6 +976,27 @@ let change_page_unknown
       `Get
     | _, _ ->
       `Post
+  and i_subpath =
+    let rec f p p' =
+      match p, p' with
+      | h :: t, h' :: t' when h = h' ->
+        f t t'
+      | [], t ->
+        Some t
+      | _ ->
+        None
+    in
+    match Eliom_request_info.get_app_path () with
+    | Some app_path ->
+      (* try to remove application path from path, client routing
+         expects paths without the former *)
+      (match f app_path i_subpath with
+       | Some i_subpath ->
+         i_subpath
+       | None ->
+         i_subpath)
+    | None ->
+      i_subpath
   in
   let%lwt () =
     route ~replace {
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -74,7 +74,8 @@ let init_client_app
     ?(ssl = false) ~hostname ?(port = 80) ~full_path () =
   Lwt_log.ign_debug_f "Eliom_client.init_client_app called.";
   (match app_path with
-   | Some app_path -> Eliom_uri.set_app_path app_path;
+   | Some app_path ->
+     Eliom_uri.set_app_path (Eliom_lib.Url.split_path app_path);
    | None -> ());
   Eliom_process.appl_name_r := Some app_name;
   Eliom_request_info.client_app_initialised := true;
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -69,14 +69,16 @@ let get_global_data () =
   | _ ->
     None
 
+let normalize_app_path p =
+  (* remove "" from beginning and end of path *)
+  let p = Eliom_lib.Url.split_path p in
+  let p = match  p with "" :: p -> p          | _ -> p in
+  match List.rev p with "" :: p -> List.rev p | _ -> p
+
 let init_client_app
     ?app_path ~app_name
     ?(ssl = false) ~hostname ?(port = 80) ~full_path () =
   Lwt_log.ign_debug_f "Eliom_client.init_client_app called.";
-  (match app_path with
-   | Some app_path ->
-     Eliom_uri.set_app_path (Eliom_lib.Url.split_path app_path);
-   | None -> ());
   Eliom_process.appl_name_r := Some app_name;
   Eliom_request_info.client_app_initialised := true;
   Eliom_process.set_sitedata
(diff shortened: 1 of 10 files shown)


vasilisp:
This is very fragile stuff, but I don't know how to do better.

src/lib/eliom_client.client.ml
```diff
@@ -70,8 +70,13 @@ let get_global_data () =
     None
 
 let init_client_app
-    ~app_name ?(ssl = false) ~hostname ?(port = 80) ~full_path () =
+    ?app_path ~app_name
```
balat: What is the difference between app_path and full_path?
Can you document this?

src/lib/eliom_client.client.mli
```diff
@@ -35,6 +35,7 @@ open Eliom_lib
     at the beginning of your html
     file, containing the full URL of your server. *)
 val init_client_app :
```
balat: Add doc for app_path and __eliom_path

src/lib/eliom_uri.shared.ml
```diff
@@ -18,8 +18,28 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 *)
 
-(*****************************************************************************)
 (* Building href *)
+
+let app_path = ref None
```
balat: If I'm right, this reference is only used on client side. Either use a `let%client` or, if you can, add a comment.

If it is server side, should probably depend on the `<site>` in config file, not be a global reference.
(We can have several Eliom applications on the same server.) 
→ put this in `sitedata`.

(Don't we already have this somewhere in sitedata?)

src/lib/eliom_uri.shared.ml
```diff
@@ -536,21 +556,22 @@ let make_post_uri_components_
...
       else None
     in
 
-
     (* absolute URL does not work behind a reverse proxy! *)
     let uri =
-      match absolute with
-      | Some proto_prefix ->
+      match absolute', !app_path with
+      | Some proto_prefix, Some app_path when absolute ->
```
balat: We should use this kind of URL only if the URL is generated client-side, by a client app.

if I understand well your code, app_path is set only by client_apps, so it's ok. But should be explained in comment.

Also please explain in comment why we are doing that.

balat:
I don't feel it so fragile, but this is weird and we must add clear comments to explain why we are doing that :)

balat:
In Eliom_uri, you need to modify the path for all non-attached coservices.
If I'm right you did it only for POST coservices?

src/lib/eliom_uri.shared.ml
vasilisp: It is only used on the client. It is not worth using Eliom syntax just for that, so I will see if I can find this info elsewhere (or else document it).

src/lib/eliom_client.client.ml
vasilisp: I don't know what `full_path` is for. To be determined.

vasilisp:
I am not sure we need to modify the path for non-POST services.

Consider a GET pathless service defined as follows:

```ocaml
let%server act_service =
  Eliom_registration.Action.create
    ~path:Eliom_service.No_path
    ~meth:(Eliom_service.Get (Eliom_parameter.int "a"))
    (fun i _ -> Lwt.return @@ Printf.printf "act_service(%d)\n%!" i)

let%client act_service = ~%act_service
```

and called as follows:

```ocaml
Eliom_client.change_page ~absolute:true ~service:act_service 22 ()
```

My latest patch (1a3f1c9bde9ff871c6d507f65ecc0697daa94e3d) leads to the path appearing in the URL twice (i.e., it was already there).

vasilisp:
@balat, I think the problem only appeared for the RPCs we do in the beginning (`init_request_rpc` in `PROJECT_NAME_mobile.eliom`), when `eliom.html` still appears in the current URL. The URLs we produce later (e.g., I tested GET pathless services) look correct. Therefore, I think I should revert 1a3f1c9 (other than its doc part).

src/lib/eliom_client.client.ml
vasilisp: I am still unsure about `full_path`. It is not documented anywhere. It is stored in `Eliom_common_base.client_process.info`, field `cpi_original_full_path`. Do you happen to remember what that one is for? If it is for the path of the app (quite plausible) I can use it.

vasilisp: `Eliom_request_info.get_csp_original_full_path` returns that field and describes it as "the full path of the URL where the client-side process is running". Is it what we need? I guess the client process can run on another URL, but not sure.

vasilisp:
Upon further reflection, I think we still need to modify the path for the other pathless services. To be done.

balat:
Indeed I detected the problem for initial requests. I don't know if it happens later.
I think it could be considered as a (theoretical) security issue to send the client path to the server.
But if you're confident that we never send it anymore, I'm ok.

balat:
We still need to fix the app_path vs full_path problem.
My understanding is that it is exactly the same (`site_dir` of sitedata).

vasilisp:
> We still need to fix the app_path vs full_path problem.
> My understanding is that it is exactly the same (site_dir of sitedata).

I merged them, and the result seems to work.

> Indeed I detected the problem for initial requests. I don't know if it happens later.
> I think it could be considered as a (theoretical) security issue to send the client path to the server.
> But if you're confident that we never send it anymore, I'm ok.

I think the problem arises because the client path appears in the current URL early in the process (and never again). Subsequent calls to pathless services seem OK. But I am not 100% sure. I can follow up with a separate PR if it turns out there is an issue.

PR #471, opened 2017-01-20
Fix absolute calls to non-attached coservices

vasilisp:
Fixes #468, which disregarded the path for all non-attached POST coservice calls with `absolute = true`.

We now store the client HTML file name (default: `eliom.html`), match the URL against that, and modify if necessary.

Needed for ocsigen/ocsigen-start#416 to work correctly.


vasilisp:
Nobody seems to have comments, so merging.

PR #475, opened 2017-01-24
Implement Eliom_client.onchangepage

vasilisp:
Permits running callbacks before a page-producing service handler is called.

See https://github.com/ocsigen/ocsigen-start/issues/417 for motivation.

### DIFF
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -643,6 +643,8 @@ let set_content_local ?offset ?fragment new_page =
 (* Function to be called for server side services: *)
 let set_content ~replace ?uri ?offset ?fragment content =
   Lwt_log.ign_debug ~section "Set content";
+  (* TODO: too early? *)
+  run_callbacks (flush_onchangepage ());
   match content with
   | None -> Lwt.return ()
   | Some content ->
@@ -909,6 +911,7 @@ let change_page (type m)
            in
            let l = ocamlify_params l in
            update_session_info l l';
+           run_callbacks (flush_onchangepage ());
            let%lwt () = f get_params post_params in
            change_url_string_protected ~replace uri;
            do_follow_up uri
--- a/src/lib/eliom_client.client.mli
+++ b/src/lib/eliom_client.client.mli
@@ -224,6 +224,13 @@ val onload : (unit -> unit) -> unit
 (** Returns a Lwt thread that waits until the next page is loaded. *)
 val lwt_onload : unit -> unit Lwt.t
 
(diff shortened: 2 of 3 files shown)


vasilisp:
Upon further reflection, we may be able to achieve the effect needed for ocsigen/ocsigen-start#418 with `onunload` (of the previous service). I am about to try. If that works, I will close this PR.

vasilisp:
For client services, the order things happen is quite unfortunate. `onunload` of the outgoing page happens after the service handler produces the incoming page. Do we change that order (small change, but one that could cause regressions), or does the `preload` solution sound cleaner?

Drup:
That seems particularly unfortunate indeed. What do you think could break ?
@balat / @vouillon Do you remember what you did it that way ?

vasilisp:
I implemented the change: https://github.com/ocsigen/eliom/commit/deacd25380465f824697aeda111b6cb63cc5789b . I tried with OS (https://github.com/ocsigen/ocsigen-start/commit/a0ef7e73cfc5e16f83915b4522ef1e00405f195a) and it seems to work. Formally a breaking change because `set_content_local` is in the `.mli`, but it is under a `(**/**)` and meant for internal use.

@Drup, I don't have any concrete breakage in mind. It should be OK. But I have had quite a bit of trouble getting the order of things right (due to the ping-pong between `Eliom_client` and `Eliom_registration`), so I am a bit reluctant to change it again :).

balat:
I'm ok with adding this. But perhaps we could find a cleaner interface (and also add other missing functions in this style?).
`onload` and `lwt_onload` are done only once, but `onpreload` is done for each new page ...
May be uniformize with `Lwt_js_events`?

vasilisp:
@balat, I think I prefer the change that would allow us to use `onunload` (https://github.com/ocsigen/eliom/commit/deacd25380465f824697aeda111b6cb63cc5789b). What do you think about that one? We wouldn't need to add `onpreload` at all.

> onload and lwt_onload are done only once, but onpreload is done for each new page ...

I was thinking that you can use the service handler to establish the appropriate `onload` handlers, because the `unload` will happen after the handler. But you can't possibly do this for `onpreload`, so we would need a cleaner way to establish permanent handlers.

vasilisp:
Renamed to `onchangepage`. Happens only once for consistency. It can be [registered recursively](https://github.com/ocsigen/ocsigen-start/pull/418/commits/185d159a703add8c2a9032e8799630b4399cf9c9#diff-0cb7ea41b0ed791982391f750b8a28a6R176) to circumvent this.

PR #478, opened 2017-01-30
eliom-distillery: convert project name to lowercase

vasilisp:
Uppercase characters cause trouble with database names, etc. These underlying issues may need to be resolved separately, but I think it is good to normalize our names anyway.

CC @balat

### DIFF
--- a/src/tools/distillery.ml
+++ b/src/tools/distillery.ml
@@ -297,15 +297,22 @@ let main () =
   in
   let dest_dir = ref None in
   let check_name name =
-    if not (Str.string_match compilation_unit_name_regexp name 0) then
-      bad "Not a valid compilation unit name: %s" name
+    let name' = String.lowercase name in
+    if name' <> name then
+      Printf.eprintf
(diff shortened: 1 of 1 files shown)


src/tools/distillery.ml
```diff
@@ -296,16 +296,30 @@ let main () =
...
-      bad "Not a valid compilation unit name: %s" name
+  let check_name =
+    let ret name =
+      if not (Str.string_match compilation_unit_name_regexp name 0) then
+        bad "Not a valid compilation unit name: %s" name
+      else
+        name
+    in
+    fun name ->
+      let name' = String.lowercase name in
```
pwbs: I think you should use `String. lowercase_ascii` if you really have to lowercase...

vasilisp: We still support `4.02.x`, and it is a pity to break compatibility just for that.

pwbs: oh...

PR #482, opened 2017-02-07
Re-raise server-side exceptions (besides sending them to the client)

jrochel:
This pull request is useful to raise awareness of exceptions which occur server-side.
Unfortunately stack traces do not work yet. They only show the innermost code position, which is rather useless. This might be related to: https://caml.inria.fr/mantis/view.php?id=6556

### DIFF
--- a/src/lib/eliom_registration.server.ml
+++ b/src/lib/eliom_registration.server.ml
@@ -990,6 +990,8 @@ module Ocaml = struct
           let%lwt res = f g p in
           Lwt.return (`Success res)
         with exc ->
+          (* prints exception and stack trace on the server console *)
+          Lwt.async (fun () -> Lwt.fail exc);
           Lwt.return (`Failure (Printexc.to_string exc))
       in
       prepare_data data


vasilisp:
I had seen this PR when you opened it, but I still can't make up my mind :).

To begin with, OCaml's standard backtraces are not all that useful with Lwt (because all you see is calls to the Lwt scheduler). You need to use Lwt's [own backtrace support](https://ocsigen.org/lwt/manual/) to get any useful info, and I am not sure how that works with the `Lwt.async` (in which case you create a new stack outside your real Lwt thread)

If we just want to log the exception, why throw it?

(Actually, I am not even sure we should be sending the serialized extension to the client. This could leak sensitive info. But that's another discussion.)

jrochel:
This will give us the origin of the exception:
```
print_endline @@ Printexc.to_string exc;
Printexc.print_backtrace stdout;
flush stdout;
```

jrochel:
Question is now: do we pass an exception handler to `make_service_handler` or do we directly log the backtrace via `Lwt_log`. I tend towards the latter.

jrochel:
UPDATE: this patch results now in useful stack traces. This is most likely due to https://github.com/ocsigen/lwt/commit/06f0b554d7844fbf5a82a5d9a34e522ce6a39897.

jrochel:
However any attempt of mine to get those stack traces by `Printexc.get_backtrace ()` and feeding it to `Lwt_log` has failed. I always get stack traces with a single entry (something with `lwt_map.ml`).

PR #484, opened 2017-03-02
Eliom_cscache.cache_list

jrochel:


vouillon:
What happens if you have several times the same key in your list?

src/lib/eliom_cscache.eliomi
```diff
@@ -40,6 +40,10 @@ type ('a, 'b) t
     cached.  *)
 val find : ('a, 'b) t -> ('a -> 'b Lwt.t) -> 'a -> 'b Lwt.t
 
+(*TODO: document*)
```
vasilisp: Please do :).

jrochel: Indeed. Done.

src/lib/eliom_cscache.eliom
```diff
@@ -47,6 +47,25 @@ let%server find cache get_data id =
...
 
+let cache_list cache get_data ids =
+  let main_thread, wakeup = Lwt.task () in
+  let tbl = Hashtbl.create 10 in
+  let thread id =
+    let%lwt () = main_thread in
+    Lwt.return @@ Hashtbl.find tbl id
+  in
+  let enqueue id =
+    try let _ = Hashtbl.find (Eliom_shared.Value.local cache ()) id in false
```
vouillon: `if Hashtbl.mem ... then ... else ...`

src/lib/eliom_cscache.eliom
```diff
@@ -47,6 +47,25 @@ let%server find cache get_data id =
...
+    with Not_found -> let () = do_cache_raw cache id (thread id) in true
+  in
+  try%lwt
+    let not_cached = List.filter enqueue ids in
+    let%lwt data = get_data not_cached in
+    List.iter (fun (id,v) -> Hashtbl.add tbl id v) data;
+    ignore [%client (List.iter (fun (id,v) -> do_cache ~%cache id v) ~%data : unit)];
+    Lwt.return @@ Lwt.wakeup wakeup ()
+  with e -> Lwt.return @@ Lwt.wakeup_exn wakeup e
```
vouillon: That should probably be:
```ocaml
Lwt.wakeup_exn wakeup e;
Lwt.fail e
```
In our case, we probably want the RPC to fail if the database query fails, rather than returning a result that triggers a lot of additional RPCs to fetch the missing data.

src/lib/eliom_cscache.eliom
```diff
@@ -59,6 +78,10 @@ let%client find cache get_data id =
   try Hashtbl.find ((Eliom_shared.Value.local cache) ()) id
   with Not_found -> load cache get_data id
 
+let find_list cache get_data ids =
+  let%lwt () = cache_list cache get_data ids in
+  Lwt_list.map_s (find cache (fun _ -> raise Not_found)) ids
```
vouillon: `assert false` instead of `raise Not_found`?

Raising `Not_found` here just does not make sense.

jrochel: Well, I thought that if `get_data` supplied by the user does not return a row for each key it could (should?) still be considered a "soft error".

jrochel:
Could not show that this resulted in fewer DB requests. Closing for now...

PR #485, opened 2017-03-06
Lwt pre allocated threads

sagotch:
Our javascript code size goes from 1673312 to 1673167 with this.

See also https://github.com/ocsigen/ocsigen-toolkit/pull/126 and https://github.com/ocsigen/ocsigen-start/pull/441

### DIFF
--- a/src/lib/eliom_bus.client.ml
+++ b/src/lib/eliom_bus.client.ml
@@ -47,7 +47,7 @@ let consume (t,u) s =
         | _ -> ());
       [%lwt raise ( e)]
   in
-  Lwt.choose [Lwt.bind t (fun _ -> Lwt.return ());t']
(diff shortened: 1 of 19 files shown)


vasilisp:
Does the reduction in allocations manifest itself in measurable RAM and CPU savings? If it is only saving 0.2KB in a 1.6MB file (within the margin of error), I would say let's just keep the pattern that we are used to :).

sagotch:
I did not measured it. Probably that it is not measurable savings. But it does not obfuscate semantic, it only add few characters in code, and it avoid some allocations.

I did not look where this optimization was used exactly, so I do not know if this tiny improvement could have any consequence on final perfs. In any case, it cannot have bad consequence. It could if the code was less readable but it is not the case.

I just find that it is a shame not to use these threads since they already are here.

Drup:
The change is cheap and the code is equally readable after, so why not ?

PR #488, opened 2017-04-12
Fix hostname used by client app

vasilisp:
The underlying issues were reported by @klakplok.

The first patch fixes a mismatch between the default hostname of the app (`defaulthostname` tag) and the one in the URL. The hostname is used as a key in the data structure that stores process info. As a result of the mismatch, reading the app name would fail, and thus all calls to app services would fail.

The second patch uses the process info sent by the server for building the initial `<base>` tag. I don't think that using host and path that appear in the URL is meaningful; we could just omit `<base>`.

### DIFF
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -121,6 +121,23 @@ let onunload_fun _ =
 
 let onbeforeunload_fun _ = run_onbeforeunload ()
 
+let set_base_url () =
+  let {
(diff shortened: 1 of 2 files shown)


pwbs:
Sadly, this makes BeSport's app fully unusable: it makes the page displayed by the webview be `https?://hostname/index.html` instead of `file:///blabla/index.html` and then `file:///blabla/eliom.html`! (and of course, `https?://hostname/index.html` is a 404...)

Issue #492, opened 2017-04-25
uncaught exception: Async: Eliom_request.Non_xml_content


pwbs:
![e](https://cloud.githubusercontent.com/assets/15014298/25376746/84510658-29a5-11e7-95fa-55dfb87d334d.png)
 
Within [this line - (eliom_request.client.ml#L276)](https://github.com/ocsigen/eliom/blob/b512a4dd4c923c08afb54ee4ecb2a2d68a396b02/src/lib/eliom_request.client.ml#L276)
```
          Lwt.return (url, Some (result r))
```
the subexpression `result r` may raise an exception. 

It's actually raised by `raise Non_xml_content` in [line 72](https://github.com/ocsigen/eliom/blob/b512a4dd4c923c08afb54ee4ecb2a2d68a396b02/src/lib/eliom_request.client.ml#L72)
but not directly, as there's an indirection: it passes by `eliom_client.client.ml` 

Inspecting the contents of `x` (the `x` in `x.XmlHttpRequest.content_xml ()`), I observed that the content_xml is absent, but `x.XmlHttpRequest.content` contains what's expected (or almost, perhaps). However `x.XmlHttpRequest.content` is of type `string`, so it doesn't help much...

vasilisp:
We are exercising this code all the time, I would guess. Is there anything special in your case? Notably, what service call does this correspond to? Is it reproducible on multiple browsers?

`content_xml` [seems to come from the `responseXML` field](https://github.com/ocsigen/js_of_ocaml/blob/b5cc12fe23fceb77977e985aefce1d0f892589b0/lib/lwt/lwt_xmlHttpRequest.ml#L53) of the raw JS object. Hard to say why that does not exist in your case.

pwbs:
I guessed indeed that this code was exercised a lot... I kinda wished it wasn't the case. haha

In our code (I mean, in BeSport's code),  it's simply coming from a
```ocaml
Eliom_client.change_page ~service:Eliom_service.reload_action () ()
```

There's likely to be something wrong with the way the navigation history is handled...

(I'm still investigating.)

pwbs:
> Is it reproducible on multiple browsers?

It's much easier to reproduce on Chrome (OSX). On Safari (OSX), it also happens but it's much harder to reproduce.

What I do on Chrome to reproduce it:

[...]

PR #493, opened 2017-04-26
hack: workaround for #492 (don't merge)

pwbs:
Well, this makes the problem described in https://github.com/ocsigen/eliom/issues/492 go away.
I'm not sure at all that it's the right approach, but it's a good lead to finding the real problem.

### DIFF
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -933,7 +933,7 @@ let change_page (type m)
            update_session_info l l';
            run_callbacks (flush_onchangepage ());
            let%lwt () = f get_params post_params in
-           change_url_string_protected ~replace uri;
+(*           change_url_string_protected ~replace uri; <---- shouldn't be done, perhaps???? *) 
            do_follow_up uri
          | None when is_client_app () ->
            Lwt.return @@ exit_to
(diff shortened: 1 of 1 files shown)


src/lib/eliom_client.client.ml
```diff
@@ -933,7 +933,7 @@ let change_page (type m)
            update_session_info l l';
            run_callbacks (flush_onchangepage ());
            let%lwt () = f get_params post_params in
-           change_url_string_protected ~replace uri;
+(*           change_url_string_protected ~replace uri; <---- shouldn't be done, perhaps???? *)
```
vasilisp: We have just called a service, and we do need to change the URL accordingly.

vasilisp:
Thanks for investigating. We are getting closer. We cannot accept either of these patches, though. We need to respect `?absolute` and produce proper URLs.

pwbs:
> Thanks for investigating. We are getting closer. We cannot accept either of these patches, though. We need to respect `?absolute` and produce proper URLs.

I agree. This patch, as it is, is not meant to be pulled, it's only elements from the investigation :)

src/lib/eliom_client.client.ml
pwbs: This (i.e., removing that line) breaks the login process! :(

PR #497, opened 2017-06-12
Always use current location to build <base>

vasilisp:
Partially reverts #488 (its `<base>` part).

The host manipulation in `Eliom_request` is an effort to generate `X-Eliom-Process-Cookies` correctly even if we access the app via an alternate hostname (i.e., one other than the server's default as per
the configuration file). Using the default hostname for `<base>` was meant as a workaround for the same issue.


vasilisp:
CC @balat @vouillon @pwbs

PR #504, opened 2017-06-16
Reason distillery

Charlesetc:
This is building off of https://github.com/ocsigen/eliom/pull/445 by @vasilisp.

It's adding the -reason flag to eliom-distillery which converts the .eliom files to reason syntax. It also adds the `%%%REASON_FLAG%%%` substitution which evaluates to `"-reason"` if -reason is passed to eliom-distillery and an empty string otherwise.

Hopefully this gets ocsigen closer to reason support - I haven't been able to get it working with ocsigen-start, but this does generate a reason ocsigen program with:
```
eliom-distillery -name project_name -template basic.ppx -reason
```

The problems current with getting ocsigen start to work are in the translation from the ocaml to reason.
* There is some [camlp4](https://github.com/ocsigen/ocsigen-start/blob/master/template.distillery/demo_pgocaml_db.ml) for postgres. This only affects 2 files. And there are some ppx forks of pgocaml but they aren't on opam
* The syntax `##.` doesn't seem to work when converting from ocaml to reason, it's used in js_of_ocaml ppx

It's possible these problems are too great right now - we might want to wait to merge the -reason flag until there's support for ocsigen start. Either way, it's a little closer to reason and ocsigen! ✈️


Drup:
I'm surprised by the `##.` issue. Are reason people aware of it ?

Charlesetc:
Hmm that's a good point about the -reason flag being specific to eliomc, etc. I'm not sure how to customize it ... maybe we could change it to REASON_ELIOMC_FLAG to be more specific, or as you say declaring support for reason in the templates makes sense. What do you think?

vasilisp:
Sorry for the delay.

The customizability of the flag and the declaration of Reason support could both be implemented via a per-template configuration/specification file. It could be useful for other features (e.g., supersede the existing`.eliom*` files), and we can discuss it later.

For now, could you work-around the use of `Filename.extension` via `Filename.check_suffix` (or otherwise)? Let's not break 4.03 compatibility just yet.

Thanks!

PR #512, opened 2017-07-25
Store dom elements,  memorize scroll position and add page transition animation

aik9508:
I add a list `history_doms` in Eliom_client. By adding dom elements into the list, one can read it from the storage when going back in history. In this way, we avoid reloading the page and make the memorization of the scroll position possible. 

I add also two functions `history_changepage_handler` and `animation_function` in Eliom_client, by customizing these two functions, we can realize the page transition animation when navigating back in history.

### DIFF
--- a/src/lib/eliom_client.client.mli
+++ b/src/lib/eliom_client.client.mli
@@ -225,14 +225,14 @@ val onload : (unit -> unit) -> unit
 val lwt_onload : unit -> unit Lwt.t
 
 
-(** [Onchangepage_event] is a record of some parameters related to
+(** [changepage_event] is a record of some parameters related to
     page changes. [in_cache] is true if the dom of the page is cached.
     [current_uri] is the uri of the current page and [target_uri] 
     is the uri of the next page. [current_id] is the state_id of
     the current page and [target_id] is the state_id of the next page.
     [target_id] is not [None] if and only if the onchangepage event
     takes place during a navigation in history. *)
-type onchangepage_event = 
+type changepage_event = 
   {in_cache:bool; 
    current_uri:string; 
    target_uri:string; 
@@ -244,7 +244,7 @@ type onchangepage_event =
 
     Just like onpreload, handlers registered with onchangepage only
     apply to the next page change. *)
-val onchangepage : (onchangepage_event -> unit Lwt.t) -> unit
+val onchangepage : (changepage_event -> unit Lwt.t) -> unit
 
 (** [onbeforeunload f] registers [f] as a handler to be called before
     changing the page the next time. If [f] returns [Some s], then we
--- a/src/lib/eliom_client.client.mli
+++ b/src/lib/eliom_client.client.mli
@@ -315,6 +315,11 @@ val init : unit -> unit
 
 val set_reload_function : (unit -> unit -> unit Lwt.t) -> unit
 
+(** Stock the dom of the current page, which will be used when going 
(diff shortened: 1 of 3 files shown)


sagotch:
You should probably use `git rebase -i` and merge the second commit into the first one.

PS: I did not read what this PR was doing, leaving it to ocsigen members.

src/lib/eliom_client.client.mli
```diff
@@ -315,6 +315,11 @@ val init : unit -> unit
 
 val set_reload_function : (unit -> unit -> unit Lwt.t) -> unit
 
+(** Stock the dom of the current page, which will be used when going 
+    back to this page through [window.history.back]
+*)
+val set_history_dom : unit -> unit
```
jrochel: I'd rather call this function `push_history_dom` or so. I know where the `set_` comes from but it doesn't make too much sense to me.

jrochel: You could add a line the comment that documents the typical use case of this function (i.e. during `onunload`).

src/lib/eliom_client.client.mli
```diff
@@ -315,6 +327,59 @@ val init : unit -> unit
 
 val set_reload_function : (unit -> unit -> unit Lwt.t) -> unit
 
+(** Store the document/body of the current page, which will be used when going 
+    back to this page in history.
```
jrochel: Elaborate on "will be used". It should be clear that this causes the page NOT to be charged again.

src/lib/eliom_client.client.mli
```diff
@@ -315,6 +327,59 @@ val init : unit -> unit
...
+    Notice that normally you need to call this function before installing the 
+    changepage handler for any specific page, in other words, before calling 
+    any [install_history_changepage_handler ()], so that the function 
+    [install_history_changepage_handler] will execute the handler function 
+    before leaving the page for the first time.
+*)
+val set_history_changepage_handler : (int -> onchangepage_event -> unit) -> unit
+
+(** [set_animation_function f] sets the animation function which is used to
+    realize the page transition on coming back in history. With the same
```
jrochel: to only when coming back but also when going forward / clicking on any link, n'est-ce pas ?

src/lib/eliom_client_core.client.ml
```diff
@@ -54,10 +54,19 @@ let create_buffer () =
...
+   current_id:int; 
+   target_id:int option}
+
+let run_onchangepage_callbacks ev handlers =
+  let rec aux = function
+  | [] -> Lwt.return_unit
+  | h :: s ->
+    let%lwt () = h ev
+    in aux s
+  in aux handlers
```
jrochel: `Lwt.iter_s`

src/lib/eliom_client.client.mli
```diff
@@ -224,12 +224,27 @@ val onload : (unit -> unit) -> unit
...
 
+
+(** [Onchangepage_event] is a record of some parameters related to
+    page changes. [in_cache] is true if the dom of the page is cached.
+    [current_uri] is the uri of the current page and [target_uri] 
+    is the uri of the next page. [current_id] is the state_id of
+    the current page and [target_id] is the state_id of the next page.
+    [target_id] is not [None] if and only if the onchangepage event
+    takes place during a navigation in history. *)
+type onchangepage_event =
```
jrochel: better name: `changepage_event`

PR #515, opened 2017-08-09
change current_state_id after page change 

aik9508:
Change current_state_id after page change when navigating through history so that the order of page change and state change will be consistent with other processes of changepage.


jrochel:
@aik9508 Please resolve the conflict.

PR #517, opened 2017-08-31
add examples for [Eliom_client.push_history_dom]

aik9508:


balat:
@jrochel do we want to merge this?

PR #527, opened 2017-10-17
Eliom_client.onarrive: execute code when page is served from DOM cache

jrochel:


src/lib/eliom_client.client.mli
```diff
@@ -227,25 +227,40 @@ val lwt_onload : unit -> unit Lwt.t
...
    target_id:int option}
 
 (** Run some code *before* the next page change, that is, before each
     call to a page-producing service handler.
-
     Just like onpreload, handlers registered with onchangepage only
     apply to the next page change. *)
 val onchangepage : (changepage_event -> unit Lwt.t) -> unit
 
+(* Run some code when arriving at a page. More specifically, this is currently
```
vasilisp: `(**`

src/lib/eliom_client.client.mli
```diff
@@ -227,25 +227,40 @@ val lwt_onload : unit -> unit Lwt.t
...
+   the only way to execute something when a page is being served from the DOM
+   cache upon navigating to a page in the browser history. If [now] is [true]
+   (default) the code is also executed right away.
```
vasilisp: Don't we need empty lines to produce different paragraphs?

src/lib/eliom_client.client.mli
```diff
@@ -245,6 +245,22 @@ type changepage_event =
...
+   This function is to be used to launch processes that are to run while a page
+   is being viewed and are not attached to an event listener belonging to a DOM
+   element. Specifically event listeners belonging to [Dom_html.window] are
+   destroyed when leaving the page and not reinstated automatically if the page
+   is served from the DOM cache.
+   Be careful to call this function at the right point in time. For instance
+   when called asynchronously it could easily be executed BEFORE the actual
+   page change! So it would be registered with the previously viewed page.
+   TODO: inform this function on which page change it belongs to, so
+   asynchronous calls are possible. *)
```
vasilisp: Can you specify that `onarrive` handlers are called for the current page only? `onarrive` does not install a handler for all pages like some other `on*` do.

PR #529, opened 2017-10-31
Provide a way to serve global data separately from the initial page

vouillon:
By not including the global data in the initial page, the global data can be cached, thus saving some computation when producing the page.

### DIFF
--- a/doc/manual-wiki/workflow-configuration.wiki
+++ b/doc/manual-wiki/workflow-configuration.wiki
@@ -396,6 +396,31 @@ The syntax is:
 
 Both attributes are optional and default to {{{false}}}.
 
+=== Serving global data through a distinct URL
+
+When an Eliom program is initiated client-side, it needs to receive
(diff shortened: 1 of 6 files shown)


Drup:
The first commit is unrelated, right?

I agree this is a good idea, the current implementation is, as the commit indicate, WIP. :p

vouillon:
The first commit is necessary for the remainder to work.
But, indeed, I should probably make a separate pull request.

vouillon:
I have put the first commit in #530.

Drup:
Is there any reason not to make that the default ? Given browser caching, this must be a net gain almost immediately.

vouillon:
It seems more robust to not make it the default: the client might receive stale global data if they change over time. (Which is probably not recommended anyway, though.)

Drup:
@vouillon Can't we load the file all the time and rely on browser caching/date headers ? 

(having state inside global data is a recipe for trouble anyways, but that's another issue)

PR #530, opened 2017-11-02
Improvement of service client functions

vouillon:
Put client functions in a client-side reference. So, several instances of a same service sent from the server to the client will always share the same client function.

### DIFF
--- a/src/lib/eliom_service_base.eliom
+++ b/src/lib/eliom_service_base.eliom
@@ -135,7 +135,11 @@ type ('get, 'post, 'meth, 'attached, 'co, 'ext, 'reg,
   (* If the service has a client-side implementation, we put the
      generating function here: *)
   mutable client_fun :
-    ('get -> 'post -> unit Lwt.t) Eliom_client_value.t option;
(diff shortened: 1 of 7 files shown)


src/lib/eliom_service_base.eliom
```diff
@@ -135,7 +135,11 @@ type ('get, 'post, 'meth, 'attached, 'co, 'ext, 'reg,
   (* If the service has a client-side implementation, we put the
      generating function here: *)
   mutable client_fun :
-    ('get -> 'post -> unit Lwt.t) Eliom_client_value.t option;
+    ('get -> 'post -> unit Lwt.t) option ref Eliom_client_value.t option;
+  (* The function is in a client-side reference, so that it is shared
+     by all occurrences of the service sent from the server.
+     For some service, we cannot create the client value immediately;
+     this is done later on using [internal_set_client_fun].  *)
```
Drup: You don't seem to be using the outermost option anymore. Can you remove it ?

vouillon: The outermost option is used in `eliom_service_base.eliom` (function `reload_action_aux`). Some services are created there, but we are not allowed to create client values at this point yet.

PR #533, opened 2017-11-06
Add configuration file options to load the application script asynchronously

vouillon:
We could already achieve this by including the application script element explicitly, but this is simpler this way.

Also, we can use the same `defer` settings for global data (#529).


doc/manual-wiki/workflow-configuration.wiki
```diff
@@ -385,6 +385,17 @@ The syntax is:
...
 * {{{<ipv6subnetmask value="ff:ff:ff:ff:ff:ff::"/>}}}
 
+=== Application script attributes
+
+One can choose to set the {{{defer}}} or {{{async}}} attributes of the
+script element referencing the application code, so that HTML
+rendering is not blocked while the script is downloaded.
+
+The syntax is:
+* {{{applicationscript defer=true async=false}}}
```
balat: With `<.../>` and double quotes for attributes?

PR #541, opened 2017-12-22
OCaml 4.06 compatibility fix (-safe-string)

ghuysmans:
Maybe some unsafe conversions are really unsafe... Help?

### DIFF
--- a/src/lib/eliom_parameter.server.ml
+++ b/src/lib/eliom_parameter.server.ml
@@ -23,7 +23,7 @@ include Eliom_parameter_base
 
 type raw_post_data =
   ((string * string) * (string * string) list) option *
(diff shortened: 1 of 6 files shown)


vasilisp:
> Maybe some unsafe conversions are really unsafe... Help?

Are you referring to some of the conversions in Ocsigenserver or to the single conversion in Eliom? I can't see how the string produced by `asprintf` could escape the scope of `out_list` and thus cause trouble.

ghuysmans:
Oh, I mixed up with what I did in Eliom, where I had the feeling that the real buffer could possibly get updated when "string references" still exist.

vasilisp:
No longer needed. Thanks anyway @ghuysmans!

PR #555, opened 2018-04-16
Prevent '_ in PPX auxilliary names

vasilisp:
This confuses our `sed` invocation in `eliomc` (used as a workaround for weakly-polymorphic variables).

Thanks @remyzorg.

### DIFF
--- a/src/ppx/ppx_eliom_utils.ml
+++ b/src/ppx/ppx_eliom_utils.ml
@@ -53,6 +53,12 @@ let file_hash loc =
     let d = 10 - (i * 6) mod 8 in
     Bytes.set o i e.[(g p lsl 8 + g (p + 1)) lsr d land 63]
   done;
+  for i = 0 to 4 do
+    (* Prevent problematic '_ pattern. This confuses our sed
+       invocation in eliomc. Simply replacing the pattern here is
+       easier than tightening the sed expression. *)
+    if Bytes.get o i = '\'' && Bytes.get o (i + 1) = '_' then Bytes.set o i 'Z'
+  done;
   Bytes.to_string o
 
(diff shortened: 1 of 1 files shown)


src/ppx/ppx_eliom_utils.ml
```diff
@@ -53,6 +53,12 @@ let file_hash loc =
     let d = 10 - (i * 6) mod 8 in
     Bytes.set o i e.[(g p lsl 8 + g (p + 1)) lsr d land 63]
   done;
+  for i = 0 to 4 do
```
remyzorg: Wasn't it possible to just remove the simple quote from input chars ?

src/ppx/ppx_eliom_utils.ml
```diff
@@ -53,6 +53,12 @@ let file_hash loc =
...
+    (* Prevent problematic '_ pattern. This confuses our sed
+       invocation in eliomc. Simply replacing the pattern here is
+       easier than tightening the sed expression. *)
+    if Bytes.get o i = '\'' && Bytes.get o i = '_' then Bytes.set o i 'Z'
```
vasilisp: I could randomize the `'Z'` with any allowed character other than `'`. But it looks like too much of a detail to worry about.

remyzorg: `Bytes.get o i = '\'' && Bytes.get o (i + 1) = '_'`

Isn't this supposed to be this condition ? (`+1` missing)

src/ppx/ppx_eliom_utils.ml
vasilisp: It would work with somewhat less-random names.

PR #556, opened 2018-04-17
Add symbolic link for fix link with versions

CelieArnaud:


vasilisp:
I merged to see what will happen. `index.html` seems to work: http://ocsigen.org/eliom/ but the others don't. Maybe try `api.html` and `manual.html`? (GitHub Pages redirects `foo` to `foo.html`.)

PR #557, opened 2018-04-18
caml_bytes_of_string -> caml_jsbytes_of_string.

sagotch:
Compile with js_of_ocaml master branch. Eventually, this will be needed, but probably not now.

Credits go to @vouillon


vasilisp:
> Eventually, this will be needed, but probably not now.

I don't understand :). `caml_jsbytes_of_string` is new, so don't we immediately need a dependency to js_of_ocaml >= dev`?

sagotch:
(It's meant to compile with OCaml `4.06.1`, and probably `4.06.1`).

PR #562, opened 2018-06-21
Client-side non-localized parameters

vasilisp:
- Fix client-side routing with NL params (not taken into account)
- Implement `Eliom_parameter.get_non_localized_get_parameters`


vouillon:
Client routine works.
We are not using Eliom_parameter.get_non_localized_get_parameters client-side, so I have not tested this function.

PR #566, opened 2018-07-13
Fix compilation against js_of_ocaml-lwt.3.2.0

Zzull:


hhugo:
The proper fix would be to depend on `lwt_log` explicitly,
like in https://github.com/ocaml/opam-repository/pull/12185/files

Zzull:
Commit amended

PR #583, opened 2018-12-14
Eliom_wrap: faster implementation

vouillon:
We no longer force anticipated minor GC collection, and we avoid duplicating the whole input value.


Drup:
Wow, slow down, can I review that one ? At least, I wouldn't some explanations on the overall approach! Also, I'm curious if you have perf numbers. :)

vouillon:
The code is rather heavily commented. Tell me if you need more.

I have not done any extensive performance measurement. But generating a fairly complex page such as the one at https://www.besport.com/ seems to take about half as much time.

Issue #596, opened 2019-03-23
can't install from opam


Et7f3:

Et7f3:
I have tested to compile with make but they are missing file like eliom.install

Et7f3:
I have tested like the README
make produce the same error
make install produce this eror
```make: *** No rule to make target 'eliom.install', needed by 'install'.  Stop.```

Et7f3:
I have tested form cygwin terminal or cmd
and the same bug happen so I show cmd syntax but is easy to convert to bash syntax or whatever (* ask me if you you *)

step to reproduce:
1) Install [cygwin 32 bits](https://cygwin.com/setup-x86.exe)

2) Install dependencies
`set root="X:\cygwin32"
setup-x86.exe --site http://cygwin.mirror.constant.com --quiet-mode --root %root% --arch x86 --disable-buggy-antivirus --local-package-dir %root% --prune-install --packages opam,libgmp-devel,libssl-devel,libgdbm-devel,libpcre-devel,zlib-devel,libgtk2.0-devel,xinit`

3) init opam with ocaml-base-compiler switch (* the default can't build dune so it can't build eliom *)
`%root%\bin\opam.exe init -n --compiler=ocaml-base-compiler --switch=ocaml-base-compiler`
 
4) clone this repo
`%root%\bin\git clone https://github.com/ocsigen/eliom.git`

5) set some environment variable (like path) or switch to cygwin terminal (mintty will do that for you)
dont forget to `eval $(opam env)`

6) cd eliom

7) make -n
`ocamlbuild -no-plugin -I src/ocamlbuild -no-links -use-ocamlfind build/build.native 1> /dev/null
ocaml pkg/build.ml manpage=false native=true native-dynlink=true`

8) execute first line
Ocamlbuild exit with 0 status code and doesn't print anything.

9) execute last line
it exit with status code 10 and print that
`ocamlfind: Package `wikidoc' not found
File "src/_tags", line 51, characters 50-65:
Warning: tag "subproject" does not expect a parameter, but is used with parameter "ppx"
File "src/_tags", line 50, characters 64-86:
Warning: tag "subproject" does not expect a parameter, but is used with parameter "ocamlbuild"
File "src/_tags", line 49, characters 60-78:
Warning: tag "subproject" does not expect a parameter, but is used with parameter "server"
File "src/_tags", line 48, characters 60-78:
Warning: tag "subproject" does not expect a parameter, but is used with parameter "client"
[...]

PR #597, opened 2019-03-28
Solve issue #596

Et7f3:
When trying to link to js_of_ocaml-ppx_deriving_json the error was permission denied so I chmod It if necessary on Cygwin.
patch for https://github.com/ocsigen/eliom/issues/596


Drup:
That doesn't seem like the right fix at all. This file should have the right permissions right from the start. eliom's build system shouldn't touch it. 

cc @hhugo this seems to be a js_of_ocaml issue.

PR #606, opened 2019-07-05
Remove Camlp4 syntax extension and adapt to OCaml 4.08

vouillon:
Seems to be buggy though, but this might be a bug in `ppx_tools`.
`ocsigen-toolkit' fails to compile:
```
File "src/widgets/ot_drawer.eliom", line 149, characters 47-49:
Error: Unbound value ~%
```


vouillon:
The issue is likely to be: ocaml-ppx/ppx_tools#75

hhugo:
Would it be time to switch to ppx_tools_versioned ?

vouillon:
@hhugo I think we have an incompatibility anyway since the signature of `Pparse.parse_interface` has changed.

vouillon:
Well, there is `Migrate_parse_tree.Parse.interface`...

facelesspanda:
noticed `eliom.syntax.predef` still depends on `lwt_camlp4` and `js_of_ocaml-camlp4`:
https://github.com/ocsigen/eliom/blob/70bb740d000172ac0d6f9aa0a47c20e0c0084284/pkg/META#L62-L72

PR #610, opened 2019-09-25
fix scroll bug: scroll_to_fragment was applied to old page

jrochel:


src/lib/eliom_registration.client.ml
```diff
@@ -204,8 +204,9 @@ module Html = Make (struct
     let reset_reload_fun = false
 
     let send ?options:_ page =
-      Eliom_client.set_content_local
-        (Eliom_content.Html.To_dom.of_element page)
+      Eliom_client_core.with_new_page ~replace:false () @@ fun () ->
```
vouillon: Do we need both this and the call to `Eliom_client_core.with_new_page` in `Eliom_client.change_page`?

src/lib/eliom_client.client.ml
```diff
@@ -1065,8 +1066,9 @@ let change_page (type m)
                (flush_onchangepage ())
            in
            with_new_page ~replace () @@ fun () ->
+           let%lwt () = f get_params post_params in
```
vouillon: I'm quite sure that swapping these two lines breaks client-side redirection. Indeed, they are implemented by calling `change_page` from function `f`, replacing the current page. If you change the url after the call to `f`, the wrong page will be replaced.

PR #612, opened 2019-10-10
fix scrolling bug: scrolling to top happens before page change

jrochel:
Server-side only redirections still are causing a bit of trouble, mangling up the history a bit.

### DIFF
--- a/src/lib/eliom_client.client.ml
+++ b/src/lib/eliom_client.client.ml
@@ -618,10 +618,11 @@ let change_url_string ~replace uri =
   Lwt_log.ign_debug_f ~section:section_page "Change url string: %s" uri;
   set_current_uri @@ fst (Url.split_fragment uri);
   if Eliom_process.history_api then begin
+    let this_page = get_this_page () in
     if replace then begin
       Opt.iter stash_reload_function !reload_function;
       Dom_html.window##.history##replaceState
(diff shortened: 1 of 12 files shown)


jrochel:
C'est en ce moment deployé sur master.besport.com. Il y a deux redirection sur la page d'admin dont une est implémentée seulement coté client.

src/lib/eliom_client.client.ml
```diff
@@ -715,10 +724,10 @@ let set_template_content ~replace ~uri ?fragment =
     run_onunload_wrapper (really_set content) cancel
 
 let set_uri ~replace ?fragment uri =
-  (* Changing url: *)
+  let advance = not replace in
```
vouillon: I don't understand this. Does this mean that, if we call `change_page` with argument `~replace:true`, then the new page will never become active?

src/lib/eliom_client_core.client.ml
```diff
@@ -762,7 +762,10 @@ let get_this_page () = match Lwt.get this_page with
     !active_page
 
 let with_new_page ?state_id ~replace () f =
-  let state_id = if replace then Some (!active_page).page_id else state_id in
+  let state_id = if replace
+    then let current_page = get_this_page () in Some current_page.page_id
```
vouillon: I'm not sure I understand this

PR #617, opened 2019-11-25
Add missing eq parameter to Eliom_shared.React.S.create

vouillon:


facelesspanda:
Well that saves me the work.

Shouldn't the server-side `FakeReact.create` take a dummy `?eq` too? akin to `FakeReact.map` and the like ?

PR #621, opened 2020-01-05
Support running PPXes from driver.

paurkedal:
This registers PPXes with the PPX driver and reimplements the command version by invoking the driver.

This will be useful for compiling Eliom projects with dune, esp. once it gets more stable support for generated rules.


vouillon:
Thanks a lot!

PR #640, opened 2020-06-23
Fix compilation with OCaml 4.11

### LINKED ISSUE #639: The ppx syntax extension (ppx_eliom_types_ex) generates code that does not compile as of OCaml 4.11
Using OCaml 4.11, Eliom fails to compile with the following error message:
```
- File "src/lib/eliom_shared_content.eliom", lines 218-283, characters 2-5:
- Error: Signature mismatch:
-        ...
-        Values do not match:
-          val string_of_alignment_baseline :
-            unit ->
-            (_[< Svg_types.alignment_baseline ], string)
-            Raw_wrapped_functions_svg.Xml.W.ft Eliom_shared.Value.t
-        is not included in
-          val string_of_alignment_baseline :
-            ([< Svg_types.alignment_baseline ], string) Xml.W.ft
-        File "lib/svg_sigs.mli", lines 1072-1073, characters 2-55:
-          Expected declaration
-        File "src/lib/eliom_shared_content.eliom", line 222, characters 8-36:
-          Actual declaration
```
I looked into it and with the help of @Drup and @octachron and the issue comes from the ppx syntax extension (specifically ppx_eliom_types_ex) which replaces the `[%shared ]` blocks by something of the form:
```
let _eliom_fragment_GH1xIy1 = Stdlib.ref None
let string_of_uri () =
  Eliom_shared.Value.create Eliom_content_core.Xml.string_of_uri
    (();
     _eliom_fragment_GH1xIy1 :=
       (Some (Eliom_syntax.client_value "" 0 : _ Eliom_client_value.t));
     (match !_eliom_fragment_GH1xIy1 with
      | Some x -> (x : _ Eliom_client_value.t)
      | None -> assert false))
```
This code hacks its way around the fact that references are monomorphic by introducing a polymorphic value of type `'a Eliom_client_value.t` and making the two arguments of `Eliom_shared.Value.create` match their types and somehow remove the monomorphic annotation.

However in https://github.com/ocaml/ocaml/pull/9546 this hack has been fixed so something else should be done instead.
If you believe there is no other way of doing this, please feel free to talk with the OCaml dev team before the release of the next beta to see what's possible.

vouillon:
Fix #639

### DIFF
--- a/src/lib/eliom_shared_content.eliom
+++ b/src/lib/eliom_shared_content.eliom
@@ -220,10 +220,14 @@ module Svg = struct
     module Xml = Xml
 
     let string_of_alignment_baseline () =
-      [%shared  Raw_wrapped_functions_svg.string_of_alignment_baseline ]
+      ([%shared  Raw_wrapped_functions_svg.string_of_alignment_baseline ] :
+         (Svg_types.alignment_baseline -> string) Eliom_shared.Value.t :>
+         ([< Svg_types.alignment_baseline ] -> string) Eliom_shared.Value.t)
 
     let string_of_big_variant () =
-      [%shared  Raw_wrapped_functions_svg.string_of_big_variant ]
+      ([%shared  Raw_wrapped_functions_svg.string_of_big_variant ] :
+         (Svg_types.big_variant -> string) Eliom_shared.Value.t :>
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks a lot! I've tested it and I can confirm this compiles on OCaml 4.11

PR #643, opened 2020-08-13
Fix compilation on bytecode architectures

glondu:


glondu:
Applied in Debian to make compilation pass on all architectures.

vouillon:
Thanks!

PR #660, opened 2020-12-16
rework cookie garbage collection

jrochel:
Depends on https://github.com/ocsigen/ocsigenserver/pull/187.

The current implementation suffers from performance issues as checking
for expired cookies requires cycling through the entire set of
persistent cookies.
This commit adds a new Ocsipersist table that maps cookie expiry dates
to cookie names, allowing for an efficient scan for potentially expired
cookies.

### DIFF
--- a/opam
+++ b/opam
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 name: "eliom"
-version: "6.12.4"
+version: "6.13.0"
 maintainer: "dev@ocsigen.org"
 authors: "dev@ocsigen.org"
 synopsis: "Client/server Web framework"
@@ -25,7 +25,7 @@ depends: [
   "lwt_log"
   "lwt_ppx"
   "tyxml" {>= "4.4.0" & < "5.0.0"}
-  "ocsigenserver" {>= "2.10"}
+  "ocsigenserver" {>= "2.17.0"}
   "ipaddr" {>= "2.1"}
   "reactiveData" {>= "0.2.1"}
   "dbm" | "sqlite3"
--- a/src/lib/eliom_common.server.ml
+++ b/src/lib/eliom_common.server.ml
@@ -1270,24 +1270,72 @@ let create_persistent_table name =
   perstables := Perstables.add name !perstables;
   Ocsipersist.open_table name
 
-let persistent_cookies_table :
-    (full_state_name * float option * timeout * perssessgrp option)
-    Ocsipersist.table Lwt.t Lazy.t =
-  lazy (create_persistent_table eliom_persistent_cookie_table)
-(* Another tables, containing the session info for each cookie *)
-(* the table contains:
-   - the expiration date (by timeout), changed at each access to the table
-     (float option) None -> no expiration
-   - the timeout for the user (float option option) None -> see global config
-     Some None -> no timeout
- *)
-(* It is lazy, because we must delay the creation of the table until
-   the initialization of eliom in case we use static linking with
-   sqlite backend ... *)
+module Persistent_cookies = struct
+  (* Another table, containing the session info for each cookie *)
+  (* the table contains:
+     - the expiration date (by timeout), changed at each access to the table
+       (float option) None -> no expiration
+     - the timeout for the user (float option option) None -> see global config
+       Some None -> no timeout
+   *)
+  (* It is lazy, because we must delay the creation of the table until
+     the initialization of eliom in case we use static linking with
+     sqlite backend ... *)
+
+  type date = float
+  type cookie = full_state_name * date option * timeout * perssessgrp option
+
+  module Cookies =
+    Ocsipersist.Table
+      (struct let name = eliom_persistent_cookie_table end)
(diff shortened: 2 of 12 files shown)


opam
```diff
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 name: "eliom"
-version: "6.12.4"
+version: "6.12.5"
```
balat, line 3: 6.13.0?

src/lib/eliom_common.server.ml
```diff
@@ -1270,24 +1270,39 @@ let create_persistent_table name =
   perstables := Perstables.add name !perstables;
   Ocsipersist.open_table name
 
-let persistent_cookies_table :
-    (full_state_name * float option * timeout * perssessgrp option)
-    Ocsipersist.table Lwt.t Lazy.t =
-  lazy (create_persistent_table eliom_persistent_cookie_table)
```
balat, line 1276: Do we still need function create_persistent_table?

src/lib/eliom_common.server.ml
```diff
@@ -1270,24 +1270,39 @@ let create_persistent_table name =
...
+      (struct let name = "eliom_persist_cookies_expiry_dates" end)
+      (Ocsipersist.Column.Float)
+      (Ocsipersist.Column.String)
+end
 
 
 (** removes the entry from all opened tables *)
 let remove_from_all_persistent_tables key =
+  (* doesn't remove entry from Persistent_cookies_expiry_dates; not a problem *)
+  Persistent_cookies.Cookies.remove key;
```
balat, line 1305: `>>= fun () ->` instead of `;`?

src/lib/server/eliommod_sessadmin.ml
```diff
@@ -211,8 +209,9 @@ let update_pers_exp full_st_name sitedata old_glob_timeout new_glob_timeout =
           | Some t when t <= now ->
               Eliommod_persess.close_persistent_state2
                 ~scope sitedata sessgrp k
-          | _ -> Ocsipersist.add table k (full_st_name2, newexp,
-                   Eliom_common.TGlobal, sessgrp) >>= Lwt_unix.yield
+          | _ ->
+            Eliom_common.Persistent_cookies.Cookies.add k
```
balat, line 213: Don't you need to update `Eliom_common.Expiry_dates` as well?
(May be it should be done in `Eliom_common.Persistent_cookies.Cookies.add`?

src/lib/eliom_common.server.ml
jrochel: It's utilised by `Eliom_state.create_persistent_table`. I'd prefer not to introduce breaking changes in this PR. Maybe later I'll port the other persistent tables to the functorial interface as well and get rid of some legacy code.

src/lib/server/eliommod_sessadmin.ml
jrochel: Done (more systematically than before).

src/lib/eliom_common.server.ml
```diff
@@ -89,14 +89,14 @@ type cookie_exp =
...
-(* 2013-03-01 From now on, cookie expire 10 years after opening the session.
+(* By default cookies expire 1 year after opening the session.
    Before, it was when the browser was closed but we think it has no sense,
    and many people do not understand why their session is closed, even if
    the session duration on server side is long.
    If you want this, you now have to set this manually.
 *)
 let default_client_cookie_exp () =
-  CESome (Unix.time () +. 315532800.)
+  CESome (Unix.gettimeofday () +. 31553280.)
```
vouillon, line 99: I don't think this is a good idea to do that now.
We should first have a mechanism to refresh cookies.
Otherwise, we will start to have many people getting unsuspectingly logged-out one year from now.

Also, essential cookies are ok with the law for most use cases in most countries.

jrochel: Oh, I had thought that cookies were already being refreshed!

vouillon: The default server-side timeout is one day (see server/eliommod_timeouts.ml).
The client-side cookies become useless once they expire server-side.

src/lib/server/eliommod_pagegen.ml
```diff
@@ -148,9 +148,7 @@ let update_cookie_table ?now sitedata (ci, sci) =
                       oldv = newc.Eliom_common.pc_value ->
                     catch
                       (fun () ->
-                        Lazy.force Eliommod_persess.persistent_cookies_table >>= fun table ->
-                        Ocsipersist.replace_if_exists
-                          table
+                        Eliom_common.Persistent_cookies.Cookies.replace_if_exists
```
vouillon, line 151: Shouldn't this insert an entry in `eliom_persist_cookies_expiry_dates`?

vouillon:
We should use `Unix.gettimeofday` instead of `Unix.time` in `Eliommod_pagegen.gen`

jrochel:
C'est fait.

PR #661, opened 2021-01-05
Eliom_client.add_header: add header to all requests

jrochel:


src/lib/eliom_client.client.mli
```diff
@@ -453,3 +453,6 @@ val log_section : Lwt_log.section
 val middleClick : Dom_html.mouseEvent Js.t -> bool
 
 type client_form_handler = Dom_html.event Js.t -> bool Lwt.t
+
+(** headers to add to each of Eliom's HTTP requests. *)
+val additional_headers : (string, string) Hashtbl.t
```
vouillon, line 458: It might be better to add a function `add_header` rather than exposing an implementation details (the fact that we use a hash table).

PR #664, opened 2021-01-12
Switch to ppxlib

vouillon:


kit-ty-kate:
Any news on this PR? Currently 4.13 does not support OMP1. I was counting on all the projects still requiring OMP1 (through ppx_tools_versioned) to migrate to ppxlib before the first beta but it depends partially on this if I have to backport the 4.13 support to OMP

vouillon:
How urgent would that be?
This PR still depends on OMP1. We need to switch from `Ast_mapper` to `Ppxlib.Ast_traverse`.

kit-ty-kate:
4.13.0~alpha2 is planned for this week, beta1 is planned for the beginning of august and the release for the beginning of september. Ideally it would be great to have eliom ready for the first beta so that people can test it, but it depends how hard it is. I could try to have a go at it if you don't have the time.

pitag-ha:
Do I understand correctly that the difficulty in replacing `Ast_mapper.mapper` by `Ppxlib.Ast_traverse` in `eliom` is that one part of the AST traverse logic is defined individually for each of the three `eliom` PPXs inside their individual `Pass` module and the rest of the AST traverse logic is defined by the `Make` functor?

src/ppx/ppx_eliom_client.ml
```diff
@@ -290,5 +292,8 @@ end
 include Make(Pass)
 
 let () =
-  Migrate_parsetree.Driver.register ~name:"ppx_eliom_client" ~args:driver_args
-    Migrate_parsetree.Versions.ocaml_408 mapper
+  Ppxlib.Driver.register_transformation
+    ~preprocess_impl:(fun str -> mapper.AM.structure mapper str)
+    ~preprocess_intf:(fun sig_ ->
```
pitag-ha, line 297: Why do you need `preprocess_impl`/`preprocess_intf`? Do the three `eliom` PPXs need to be applied before other PPXs are applied?

PR #666, opened 2021-02-16
[opam] incompatibility with js_of_ocaml.3.9.0

jrochel:
see https://github.com/ocaml/opam-repository/pull/18160/files


hhugo:
I suspect the fix would have resulted in a smaller diff :)

PR #667, opened 2021-02-16
Fix for jsoo 3.9

hhugo:
cc @jrochel @kit-ty-kate


hhugo:
don't we have any CI for eliom ?

PR #671, opened 2021-03-23
Eliom_state: add functions to get current states for all scopes

balat:
Also merge https://github.com/ocsigen/ocsigen-start/pull/622

### DIFF
--- a/opam
+++ b/opam
@@ -1,6 +1,6 @@
 opam-version: "2.0"
 name: "eliom"
-version: "6.15.0"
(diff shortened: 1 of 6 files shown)


balat:
@vouillon I replaced the functions by more generic ones that work for all scopes.
(This changes the interface)

vouillon:
This is an incompatible change, which breaks `ocsigen-start`. Shall we preserve compatibility for now, or bump the version number to `7`?

PR #675, opened 2021-04-16
Eliom_registration.transform_global_app_uri

jrochel:
Add function to transform the global_app URI for some obscure reason.


jrochel:
@balat @vouillon ping

PR #689, opened 2021-06-13
UNFINISHED! Hash session cookie value in si to avoid session id leaks

balat:
We now use hash as key in session tables
Warning: will close all existing sessions!!!

Remains to be done:
 - find good hash function
 - do we want to keep the initial value to keep functions Eliom_state.set_*_cookie_exp_date?
 in that case I think there is no other way than keeping the old value somewhere (instead of a_cookie_was_sent_by_browser)
 (in that case we'll need to adapt compute_new_ri_cookies too? (exp in the past?))
 - use an abstract type for hashed values (?)
 - transition period to avoid closing all sessions?

### DIFF
--- a/src/lib/server/eliommod_cookies.ml
+++ b/src/lib/server/eliommod_cookies.ml
@@ -60,7 +60,7 @@ let get_cookie_info
               Eliommod_sessiongroups.Serv.remove sessgrpnode;
               ((Eliom_common.Full_state_name_table.add
                   name
-                  (Some value          (* value sent by the browser *),
+                  (true          (* some value was sent by the browser *),
                    ref Eliom_common.SCData_session_expired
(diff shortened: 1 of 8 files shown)


src/lib/server/eliommod_cookies.ml
```diff
@@ -274,7 +274,9 @@ let compute_session_cookies_to_send
         | Eliom_common.SCData_session_expired -> None
         | Eliom_common.SC c ->
             let v = c.Eliom_common.sc_set_value in
+(*
             c.Eliom_common.sc_set_value <- `Used;
```
vouillon, line 278: Reverted since the tab cookies to send can be computed twice: once for the headers, once for the HTML page (to set a JavaScript variable)

PR #713, opened 2021-09-08
Rework Eliom_client and Eliom_client_core

jrochel:
Cosmetics and code improvements, no functional changes.
There is a minor interface change that doesn't justify a major version change in my opinion: Eliom_client.log_section → Eliom_client.section


jrochel:
@vouillon @balat ping

PR #724, opened 2021-11-25
[Eliom_client] exit_to and change_page can now open in new tab

balat:


src/lib/eliom_request.client.ml
```diff
@@ -75,12 +75,21 @@ let xml_result x =
 let string_result x = x.XmlHttpRequest.content
 
 (*TODO: use Url.Current.set *)
-let redirect_get url = Dom_html.window##.location##.href := Js.string url
+let redirect_get ?window_name ?window_features url =
```
Zzull, line 78: Can you merge those two arguments into one (in a couple for instance) since calling this function with `window_name` at `None` discards any value provided for `window_features`?

src/lib/eliom_client.client.mli
```diff
@@ -135,9 +146,10 @@ val exit_to :
   ?keep_get_na_params:bool ->
   'a -> 'b -> unit
 
-(** Loads an Eliom service in a window (cf. Javascript's [window.open]). *)
+(** Loads an Eliom service in a window (cf. Javascript's [window.open]).
+*)
 val window_open :
-  window_name:Js.js_string Js.t ->
+  window_name:Js.js_string Js.t  ->
```
Zzull, line 152: Looks like an unwanted change

src/lib/eliom_request.client.ml
balat: Yes but this would break uniformity/compatibility with previous versions ...

src/lib/eliom_client.client.mli
```diff
@@ -119,8 +124,14 @@ val call_ocaml_service :
 
 (** Stop current program and load a new page.  Note that for string arguments,
     sole line feed or sole carriage return characters are substituted by the
-    string ["\r\n"]. *)
+    string ["\r\n"].
+    If [window_name] is specified (for example ["_blank"]), open in new window.
```
vouillon, line 128: We might want to document that this does not work in a mobile app

PR #733, opened 2021-12-23
[cookies] no longer log cookie IDs but only hashes

jrochel:
also: render cookie GC less verbose

### DIFF
--- a/src/lib/eliom_common.server.ml
+++ b/src/lib/eliom_common.server.ml
@@ -1389,10 +1389,13 @@ module Persistent_cookies = struct
 
   let garbage_collect ~section gc_cookie =
     let now = Unix.time () in
-    Expiry_dates.iter ~lt:now @@ fun date cookies ->
+    Expiry_dates.iter ~lt:now @@ fun date cookies_str ->
+      let cookies = String.split_on_char ',' cookies_str in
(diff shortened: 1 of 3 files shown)


src/lib/eliom_common.server.ml
```diff
@@ -1389,10 +1389,13 @@ module Persistent_cookies = struct
 
   let garbage_collect ~section gc_cookie =
     let now = Unix.time () in
-    Expiry_dates.iter ~lt:now @@ fun date cookies ->
-      Lwt_log.ign_notice_f ~section "potentially expired cookies %.0f: %s"
-                                    date cookies;
-      Lwt_list.iter_s gc_cookie (String.split_on_char ',' cookies) >>= fun _ ->
+    Expiry_dates.iter ~lt:now @@ fun date cookies_str ->
+      let cookies = String.split_on_char ',' cookies_str in
+      let hash c = Hashed_cookies.to_string @@ Hashed_cookies.hash c in
```
vouillon, line 1394: The function `Hashed_cookies.hash` is not a good hash function. Here, I think this is just the identity, since `c` does not ends with character `H`.

jrochel: https://github.com/ocsigen/eliom/pull/738

PR #735, opened 2021-12-29
configuration option for omitting cookie creation

jrochel:
Adds a configuration option that allows for the omission of the storage
of any persistent information on the server for any requests that
satisfy all of the given conditions. Conditions can be expressed as
PCREs applied to any HTTP header of the request. This can be used to
prevent inundating the cookie table by cookies generated for crawlers
(that usually do not store cookies on their end anyway).

Documentation: https://github.com/ocsigen/eliom/pull/734

### DIFF
--- a/src/lib/eliom_common.server.ml
+++ b/src/lib/eliom_common.server.ml
@@ -349,6 +349,8 @@ type node_info = {
 
 module Hier_set = String.Set
 
+type omitpersistentstorage_rule = HeaderRule of Ocsigen_header.Name.t * Re.re
+
 type server_params =
   {sp_request: Ocsigen_extensions.request;
    sp_si: sess_info;
@@ -495,6 +497,7 @@ and sitedata =
    mutable html_content_type : string option;
    mutable ignored_get_params : (string * Re.re) list;
    mutable ignored_post_params : (string * Re.re) list;
+   mutable omitpersistentstorage : omitpersistentstorage_rule list option;
   }
 
 and dlist_ip_table = (page_table ref * page_table_key, na_key_serv)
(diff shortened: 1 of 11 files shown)


vouillon:
Is comet still working? (Doesn't this make the page reload constantly?)

vouillon:
What about Eliom references? If there is no longer any cookie associated to them in the database, will they ever be garbage collected?

jrochel:
> Is comet still working? (Doesn't this make the page reload constantly?)

I observe no visible change (aside from the database remaining unchanged). Everything works as previously. After all `Os_comet` seems to use volatile references exclusively.

jrochel:
> What about Eliom references? If there is no longer any cookie associated to them in the database, will they ever be garbage collected?

Indeed, I failed to omit storage for `Eliom_references`, which thus, as you point out, would never be garbage collected. I fixed that by omitting storage for `Eliom_references` too.

src/lib/eliom_common.server.ml
```diff
@@ -1338,6 +1341,11 @@ module Persistent_tables = struct
       ) !functorial_tables
     >>= fun functorial_counts ->
     Lwt.return @@ polymorphic_counts @ functorial_counts
+
+  let omit_storage_key = Lwt.new_key ()
+  let do_omitting_storage b = Lwt.with_value omit_storage_key (Some b)
```
vouillon, line 1346: Currently, request-specific information is stored using key `Eliom_common.sp_key` in a value of type `Eliom_common.server_params`. Could we just extend this type?

src/lib/eliom_reference.server.ml
```diff
@@ -176,20 +176,20 @@ let eref_from_fun_ ~ext ~scope ?secure ?persistent f : 'a eref =
...
       end
     | `Site ->
       begin
         match persistent with
-          | None -> (Volatile.eref_from_fun_ ~ext ~scope ?secure f :> _ eref)
-          | Some name ->
+          | Some name when not @@ Eliom_common.Persistent_tables.check_if_omitting_storage () ->
 (*VVV!!! ??? CHECK! *)
               (f, ext, Ocsiper_sit (Ocsipersist.open_table name))
+          | _ -> (Volatile.eref_from_fun_ ~ext ~scope ?secure f :> _ eref)
```
vouillon, line 192: I'm not sure about this, since references are usually created at toplevel, in a context where `Eliom_common.Persistent_tables.check_if_omitting_storage ()` is false. For instance:
```
let%server has_followed =
  Eliom_reference.eref ~persistent:"Bs_ref__has_followed"
    ~scope:Eliom_common.default_group_scope false
```

jrochel: In this specific case this is not a problem, since for `default_group_scope` the data is stored lazily :
```
  let eref_from_fun_ ~ext ~scope ?secure f : 'a eref =
    f, ext, match scope with
      | `Request -> Req (Polytables.make_key ())
      | `Global -> Ref (ref (Lazy.from_fun f))
      | `Site -> Sit (Polytables.make_key ())
      | (#Eliom_common.user_scope as scope) ->
        Vol (lazy (create_volatile_table ~scope ?secure ()))
```

For the more general case I'm wondering whether the situation you're mentioning is even possible. Will Eliom ever store persistent data associated with some cookie C from some context where C is not known? Investigating...

vouillon: Why are you quoting `Volatile.eref_from_fun_`? Aren't we going to line 199?
```ocaml
          (f, ext, Per (create_persistent_table ~scope ?secure name))
```

src/lib/eliom_common.server.ml
jrochel: There is an even better solution. We can simply evaluate the headers from the server parameters.

src/lib/eliom_reference.server.ml
jrochel: Indeed, my approach was dumb. I've taken a new approach which seems to work well. I only intervene at the last moment before the data is actually written.

PR #737, opened 2022-01-05
[cosmetics] refactor some code; reduce warnings during compilation

jrochel:


src/lib/eliom_common.server.ml
```diff
@@ -272,16 +272,18 @@ type 'a servicecookiestable = 'a servicecookiestablecontent SessionCookies.t
...
-   timeout ref             (* user timeout *) *
-   cookie_level sessgrp ref   (* session group *) *
-   string Ocsigen_cache.Dlist.node (* session group node *))
+module Data_cookie = struct
+  (* non persistent cookies for in-memory data *)
+  type t = {
+    full_state_name : full_state_name;
+    expiry : float option ref;
+    timeout : timeout ref;
+    session_group : cookie_level sessgrp ref;
```
vouillon, line 281: Could we use mutable fields instead?

src/lib/server/eliommod_cookies.ml
```diff
@@ -42,11 +50,23 @@ module Persistent_cookies = struct
      the initialization of eliom in case we use static linking with
      sqlite backend ... *)
 
-  type date = float
-  type cookie = Eliom_common.full_state_name * date option * Eliom_common.timeout * Eliom_common.perssessgrp option
+  type stored_cookie = Eliom_common.full_state_name * date option * Eliom_common.timeout * Eliom_common.perssessgrp option
 
   module Ocsipersist = Ocsipersist.Functorial
 
+  (* for backwards compatibility we cannot apply [Marshal] directly to [cookie];
```
vouillon, line 57: Are you sure you changed the data representation?

src/lib/eliom_common.server.ml
jrochel: Unfortunately not. Not easily in any case. In `src/lib/server/eliommod_cookies.ml` for instance we pass on these references to other records:
```
                        {Eliom_common.dc_hvalue= hvalue       (* value *);
                         Eliom_common.dc_set_value= None;
                         Eliom_common.dc_timeout= timeout
                          (* user timeout ref *);
                         [...]
                         Eliom_common.dc_session_group= session_group;
                         Eliom_common.dc_session_group_node= session_group_node}
```

These references are then modified in other contexts :
```
src/lib/eliom_state.server.ml:    c.Eliom_common.dc_session_group := n;
src/lib/server/eliommod_datasess.ml:            c.Eliom_common.dc_session_group := fullsessgrp
```

I'm open to change these into mutable fields, but I couldn't do so without understanding the code much better than I currently do. Maybe after some more (more trivial) refactoring I will be able to.

vouillon: Ok

src/lib/server/eliommod_cookies.ml
jrochel: I just assumed that I did. But indeed, according to my tests this was not the case. I have removed the transformation code.

jrochel:
@vouillon @balat This code has been relatively well tested. It should be semantically neutral in any case. I think it can be merged.

src/lib/eliom_request_info.server.mli
```diff
@@ -233,10 +233,6 @@ val get_persistent_cookies :
 val get_data_cookies :
   unit -> string Eliom_common.Full_state_name_table.t
 
-
-(** Returns the http error code of the request before Eliom was called *)
-val get_previous_extension_error_code :unit -> int
```
balat, line 238: Are you sure that nobody will need that?

src/lib/server/eliommod_pagegen.ml
```diff
@@ -226,17 +227,19 @@ let execute
...
-    (fun e -> handle_site_exn e info sitedata)
-  >>= fun result ->
-  update_cookie_table ~now sitedata
-    ((service_cookies_info, data_cookies_info, pers_cookies_info), secure_ci)
-  >>= fun () ->
-
-  update_cookie_table ~now sitedata
-    ((service_tab_cookies_info, data_tab_cookies_info, pers_tab_cookies_info), secure_ci_tab)
-  >>= fun () ->
+  let* result =
```
balat, line 223: Did we choose between `let*` and `lwt`?

src/lib/eliom_request_info.server.mli
jrochel: Oh, sorry. For some reason I thought that the function was not visible to the outside world. Fixed.

src/lib/server/eliommod_pagegen.ml
jrochel: Oh, for some reason I thought that we didn't use `lwt_ppx` in eliom.
Here is a PR replacing all `let*` by `let%lwt` https://github.com/ocsigen/eliom/pull/741.

PR #738, opened 2022-01-13
[cookies] fix: no longer log cookie IDs but only hashes

jrochel:
The original solution still printed cookie strings for old cookies
(which do not end on 'H').

Fixes 1206469b8001170587b21c1db483a5b75a10c440.

### DIFF
--- a/src/lib/eliom_common.server.ml
+++ b/src/lib/eliom_common.server.ml
@@ -105,19 +105,21 @@ module SessionCookies =
 (* keys in tables are hashes of cookie values *)
 module Hashed_cookies : sig
     type t
+    val sha256 : string -> string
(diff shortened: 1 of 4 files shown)


vouillon:
I would prefer it if you defined a separate function.
For instance: `val sha256 : string -> string`

jrochel:
> I would prefer it if you defined a separate function. For instance: `val sha256 : string -> string`

Done.

vouillon:
You could use function `sha256` in function `hash`.

jrochel:
> You could use function `sha256` in function `hash`.

Of course. Completely forgot about that.

jrochel:
Done.

PR #762, opened 2023-08-08
Upgrade to OCaml 5.1

kit-ty-kate:
I'm not sure what you're using to support several versions of the compiler but feel free to modify this PR in any case


vouillon:
Thanks! I think we are going to use `ppx_optcomp` to preserve compatibility with previous versions of the OCaml compiler. See #763.

PR #766, opened 2023-10-06
React fixes

vouillon:
Now that weak references are properly implemented by Js_of_ocaml, we need to make sure that signals and events do not get garbage collected too early.


balat:
Great! Thanks

PR #803, opened 2024-09-16
Avoid 'assert false' when looking up module .cmo

Julow:
This error was reachable when also hitting this error:

    Error: Error: Inferred type of injected ident not found. You need to
           regenerate services/.services.objs/byte/game.cmo.

### DIFF
--- a/src/ppx/ppx_eliom_client.ml
+++ b/src/ppx/ppx_eliom_client.ml
@@ -88,9 +88,10 @@ module Pass = struct
   let find_fragment loc id =
     if Mli.exists ()
     then Mli.find_fragment id
-    else if Cmo.exists ()
-    then Cmo.find_fragment loc
-    else [%type: _]
+    else
+      match if Cmo.exists () then Cmo.find_fragment loc else None with
+      | Some typ -> typ
+      | None -> [%type: _]
 
   let register_client_closures client_value_datas =
     let registrations =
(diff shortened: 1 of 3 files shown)


Julow:
Also, I'm hitting this error while trying to move `.eliom` and `dune` files into a subdirectory. Any idea how I could troubleshoot this error?

vouillon:
We have rules like:
```
(rule (target wup_page.ml) (deps ../wup_page.eliom)
  (action
    (with-stdout-to %{target}
      (chdir .. (run tools/eliom_ppx_client.exe --as-pp -server-cmo %{cmo:../wup_page} --impl wup_page.eliom)))))
```
Is it possible that the dependency on `%{cmo:../wup_page}` (the  server-side `cmo` file) is missing in your case?

Julow:
Thanks for the help! The rule that generates the error is this one:
```dune
(rule (target game.ml) (deps ../services/game.eliom)
  (action
    (with-stdout-to %{target}
      (chdir .. (run ../tools/eliom_ppx_client.exe --as-pp -server-cmo %{cmo:../services/game} --impl %{deps})))))
```

That `%{cmo:...}` is substituted with a correct path. The invocation is like this, perhaps the PPX doesn't like the directory structure ?
```
$ (cd _build/default/src && ../tools/eliom_ppx_client.exe --as-pp -server-cmo services/.services.objs/byte/game.cmo --impl services/game.eliom) > _build/default/src/client/game.ml
```

vouillon:
Maybe the issue is that `eliom_ppx_client` and `eliom_ppx_server` are called with different file path?
We are hashing the whole file name to generate unique ids, but we should probably hash just the base file name.
https://github.com/ocsigen/eliom/blob/b264bf98e9e96354f19e6623ce4135367c362e13/src/ppx/ppx_eliom_utils.ml#L53

PR #809, opened 2024-12-02
Fix eliom client with jsoo 5.9.0

hhugo:
It would be nice to add a simple test in the codebase that checks that generating a js file including eliom.client works.
This test could be run by opam when releasing new packages and would allow us to spot such breakage before releasing into opam.

I did a spring cleaning in jsoo runtime functions and drop caml_new_string. I will try to reintroduce it temporarily with a deprecated warning instead.


balat (approved):

balat:
Thank you Hugo!

PR #810, opened 2024-12-18
ppx: Remove sensitivity to paths passed on the command line

Julow:
eliom_ppx_client and eliom_ppx_server could be passed different paths to the same files, as Dune execute commands from different current directories.

However, this might create collisions if two file have the same name.

This was suggested by @vouillon in https://github.com/ocsigen/eliom/pull/803


Julow:
This does fix the problem that I described in https://github.com/ocsigen/eliom/pull/803. A reproduction is found by running `dune build dist` in https://github.com/Julow/ocsigen-tictactoe/tree/move-to-src

balat:
@vouillon are you happy with this?

vouillon (approved):

PR #814, opened 2025-02-20
Add upper bound for js-of-ocaml compiler in dune project

n-osborne:
Self-explanatory. Should fix CI and #813 as opam file is generated.


vouillon:
Thanks!

PR #815, opened 2025-02-20
Remove `lwt_ppx` dependency

n-osborne:
This PR proposes to get rid of the `%lwt` syntax from `lwt_ppx` in favor of the
`Lwt_syntax`.

It is based on #814, please consider only the last commit.

This is join work with @Julow (who wrote the tool making the automatic changes)

Most of the changes has been done automatically using the prototype tool
[here](https://github.com/Julow/lwt_ppx_to_let_syntax). This is the reason why
the PR is one big commit.

This is a first step toward moving from Lwt to effect-based concurrency. The
idea is to homogenize the syntax in order to have less cases to handle (again
by an automatic tool).

This is also an independent contribution, as it removes one dependency and one
preprocessor. Which can be considered a win by itself.

### DIFF
--- a/src/lib/eliom_form.eliom
+++ b/src/lib/eliom_form.eliom
@@ -20,6 +20,7 @@
 
 open%shared Js_of_ocaml
 [%%client.start]
+open Lwt.Syntax
 
 let read_params form y =
   Eliom_parameter.reconstruct_params_form (Form.form_elements form) y
@@ -35,7 +36,7 @@ let iter_contents y ev f =
   Js.Opt.case (Dom_html.CoerceTo.form target) fls @@ fun target ->
   match read_params target y with
   | Some v ->
-      let%lwt () = f v in
+      let* () = f v in
       Lwt.return_true
   | None -> !error_handler ()
 
--- a/src/lib/eliom_shared.eliom
+++ b/src/lib/eliom_shared.eliom
@@ -19,11 +19,13 @@
  *)
 
 [%%shared
+open Lwt.Syntax
+
 (* put this in Lwt_react? Find a better name? *)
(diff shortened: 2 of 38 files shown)


src/lib/eliom_form.eliom
```diff
@@ -34,9 +35,9 @@ let iter_contents y ev f =
   Js.Opt.case ev##.target fls @@ fun target ->
   Js.Opt.case (Dom_html.CoerceTo.form target) fls @@ fun target ->
   match read_params target y with
-  | Some v ->
-      let%lwt () = f v in
-      Lwt.return_true
+  | Some v -> let* () = f v in
+
```
balat, line 39: ?

src/lib/eliom_shared.eliom
```diff
@@ -19,14 +19,17 @@
...
 (* put this in Lwt_react? Find a better name? *)
 let to_signal ~init ?eq (th : 'a React.S.t Lwt.t) : 'a React.S.t =
   let s, set = React.S.create ?eq init in
   Lwt.async (fun () ->
-    let%lwt ss = th in
-    let effectful_signal = React.S.map (fun v -> set v) ss in
-    ignore (React.S.retain s (fun () -> ignore effectful_signal));
-    Lwt.return_unit);
+      let* ss = th in
+
```
balat, line 29: There seems to be a white line after many let*. And indentation is different. Did you ocamlformat everything?

n-osborne: Sadly ocamlformat doesn't seem to remove those newlines. I've updated by hand.

balat (approved):

PR #821, opened 2025-04-10
Eliom_lib: Remove deprecated logging functions

Julow:
These functions were deprecated for more than 10 years, since the introduction of `Lwt_log`: https://github.com/ocsigen/eliom/pull/117

### DIFF
--- a/src/lib/eliom_lib.client.ml
+++ b/src/lib/eliom_lib.client.ml
@@ -101,14 +101,6 @@ let _ =
         (Js.string (Printexc.to_string exn))
         exn
 
-(* Deprecated ON *)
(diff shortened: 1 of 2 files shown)


smorimoto (approved):

hhugo:
It would have been better to deprecate them with an attributes `[@@deprecated "use blah instead"]` so that user have some hints about the migration.

Julow:
These functions required a little amount of work to be ported to `Logs` and that was my main motivation for removing them.
I don't mind adding them back with a deprecated alert but do you think they are used somewhere ? I can't find any qualified occurrences with Sherlocode.

Julow:
Added in https://github.com/ocsigen/eliom/pull/827

PR #818, opened 2025-04-07
Remove uses of Ocsigen_lib Option.Lwt and Dlist.lwt_fold

Julow:
These functions have to be removed to be able to switch to direct-style concurrency.

These are used in very few places and removing them doesn't make the code worse to read.

PR #826, opened 2025-04-22
Fix incorrect rewrite introduced in #818

smorimoto:


smorimoto:
Merge #825 first.

src/lib/server/eliommod_cookies.ml
```diff
@@ -104,14 +104,14 @@ module Persistent_cookies = struct
     match expiry with
     | Some t ->
         Expiry_dates.add_cookie t cookie >>= fun _ -> Cookies.add cookie content
-    | None -> Lwt.return_unit
+    | None -> Cookies.add cookie content
```
Julow, line 107: I think it would be better if the code was not duplicated: `( match expiry with ... ) >>= fun _ -> Cookies.add ...`.
This will be rewritten to `let _ = match expiry with ... in Cookies.add ...` after the direct-style transition, which looks the most readable.

smorimoto: Indeed. Fixed!

vouillon (approved):

PR #828, opened 2025-04-22
Do not depend on lwt_log client-side

vouillon:
One should use Lwt_log_js and not Lwt_log.


eliom.opam
```diff
@@ -32,7 +32,6 @@ depends: [
   "lwt_log"
   "tyxml" {>= "4.6" & < "4.7"}
   "ocsigenserver" {>= "6.0" & < "7.0"}
-  "lwt_log"
```
vouillon, line 35: This line is duplicated

smorimoto (approved):

PR #829, opened 2025-04-22
Fix #810: keep hashing the filename

vouillon:
Since we only keep the first 36 bits of the string, we get collisions pretty easily if we use the filename directly.


src/ppx/ppx_eliom_utils.ml
```diff
@@ -283,7 +283,7 @@ module Cmo = struct
            ; ev_kind = Event_after ty
            ; _ } ->
              if pos_cnum' = pos_cnum + 1
-             then Hashtbl.add events (Filename.basename pos_fname, pos_cnum) ty
```
vouillon, line 286: This is only used internally, so we can keep the full path here.

Julow (approved):
Makes sense! It would be nice to have a testsuite for the ppx to avoid these in the future. I tested https://github.com/ocsigen/eliom/pull/810 in an ocsigen application that was too small to encounter a collision.

Issue #812, opened 2025-01-07
missing primitives bigstringaf_blit_from_bytes and bigstringaf_blit_to_bytes


njaros:
Hello, I'm trying to understand ocsigen framework thanks to the tutorials but the tutorial just doesn't work :(

I'm on this tutorial page `https://ocsigen.org/tuto/8.0/manual/application`, the Graffiti exemple.

problems encountered : 
1- the eliom-distillery template "client-server.basic" doesn't exists
2- the codes `Dom_html.window##alert` or `let%client _ = Eliom_lib.alert "Hello!"` compiles but does nothing.
3- on compilation I have missing primitives : bigstringaf_blit_from_bytes and bigstringaf_blit_to_bytes which print fatal errors on my browser console.

njaros:
I found a fix on github projects with Ocsigen:

on Makefile I had to modify those lines :
JS_OF_ELIOM := js_of_eliom -jsopt +bigstringaf/runtime.js -ppx
OCSIGENSERVER := DYLD_LIBRARY_PATH=$(shell opam var lib)/stublibs ocsigenserver

and it works

PR #831, opened 2025-04-22
Js_of_eliom: use ocamlfind to get the JavaScript runtime files

vouillon:
The JavaScript runtime files were not automatically used.

This fixes the `basic.ppx` template, which now has a dependency on package `bigstringaf` through `uri`.


smorimoto (approved):

hhugo:
Should fix https://github.com/ocsigen/eliom/issues/812

PR #837, opened 2025-06-03
Remove uses of 'Lwt_preemptive'

Julow:
`Lwt_preemptive` was not used by Eliom but code was here to print stats about the number of busy threads, which is removed.

This goes with https://github.com/ocsigen/ocsigenserver/pull/264


Julow (commented):
https://github.com/ocsigen/ocsigenserver/pull/264 was merged, this should be now.

PR #844, opened 2025-07-18
client: Direct-style API for `Eliom_comet` and `Eliom_bus`

Julow:
This changes the client-side API of `Eliom_comet` and `Eliom_bus` to be more compatible with direct-style programming:

- Add a new callback-based API, `Eliom_bus.register`, which avoids using `Lwt_stream` internally.
  `Eliom_bus.stream` is re-implemented against the register API.
  `Eliom_bus.original_stream` is not re-implemented yet.

- `Eliom_comet.Channel` is an abstract type instead of a `Lwt_stream`.
  `Lwt_stream` is difficult to migrate to other concurrency libraries because of its vast API and its many usecases.
  It's also callback-based but it's made to be easy to construct a `Lwt_stream` on top of it.

This might increase performances as well.

This was successfully tested against a small application that uses these APIs: https://github.com/Julow/ocsigen-tictactoe/pull/1

Some work remains to be done:

- [ ] API to unregister callbacks. It was common to cancel a listening loop on a channel using `Lwt.cancel` or `Lwt.pick`.
- [ ] Simplify the implementation of `Eliom_bus`. For example, the callback list could possibly be removed and implemented directly against `Eliom_comet.register`. This change also opens the door for performance improvements and code simplifications in `Eliom_comet`.

### DIFF
--- a/src/lib/eliom_comet.client.mli
+++ b/src/lib/eliom_comet.client.mli
@@ -110,22 +110,26 @@ module Configuration : sig
 end
 
 module Channel : sig
-  type 'a t = 'a Lwt_stream.t
+  type 'a t
+
+  val register : 'a t -> ('a option -> unit Lwt.t) -> unit
+  (** [register chan callback] registers a callback to be called for new messages
+    from the server. The callback receives [Some data] for each new messages
+    from the server and [None] when the server closes the channel or an error
+    occurs. Not thread-safe. *)
 end
 
 (**/**)
 
-val register :
+val register_wrapped :
    ?wake:bool
   -> 'a Eliom_comet_base.wrapped_channel
   -> ('a option -> unit Lwt.t)
-  -> unit
-(** [register ~wake chan callback] registers a callback to be called for new
-    messages from the server. If wake is false, the registration of the channel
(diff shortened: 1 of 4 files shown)


Julow:
In the last 4 commits, I improve the code a lot and added the unregistration API. The test-case is uptodate: https://github.com/Julow/ocsigen-tictactoe/pull/1

src/lib/eliom_comet.client.mli
```diff
@@ -115,17 +110,41 @@ module Configuration : sig
...
+      stop receiving new messages. No-op if the callback was unregistered before. *)
+
+  (**/**)
+
+  val wake : 'a t -> unit
+  (** Activate the handling loop, making sure the channel can receive messages.
+      No request will be sent. *)
+
+  val close : 'a t -> unit
+  (** [close c] closes the channel c. This function should be only use
```
balat, line 136: should only be used?

Julow: I don't think this docstring is true anymore, there is no way to close the channel without also using `Eliom_bus`.
I made the `close` function public and fixed a potential memory leak in my last commit.

PR #857, opened 2026-04-22
Proposal for an "ocsigen" OCamlformat profile

Julow:
I open this PR to ask for comments. I'd like to create an "ocsigen" profile in OCamlformat (the name can be debated) but it can't match exactly the profile used in ocsigen projects.

The reason is the `ocp-indent-compat` option, which is very unstable and is hard to maintain within OCamlformat. This option doesn't do what it says and causes formatting changes after each releases of OCamlformat. Internally, it just turns on many formatting specific to the janestreet profile.
It's not reasonable to use it in an other profile because it causes bugs in combination with other options that are very hard to debug.

This is the output of the profile I propose for the ocsigen projects.

Advantages:

- This reduces the indentation of argument lists and of long `(fun .. -> ..)`, which is in the spirit of the previous profile.
- Formatting of comments is enabled. This might mess up the custom doc format, so we can revert this.
- The closing `*)` of copyright headers is aligned with the rest of the comment.
- It removes 1500 lines of code !

Other changes:

- Arguments are no longer formatted one per line. This removes many lines but might be harder to edit.
- I removed the options that were equal to the default.


Julow:
In the last commit, I add an option to see more easily the end of "big" expression. It makes both reading and editing easier but it re-add the 1500 removed lines. I explain it better in the commit message.

balat:
I'm ok with these changes, except the indentation of comments. I recently added
```
parse-docstrings = false
wrap-comments = false
```
to the default `.ocamlformat`.

The ocp-indent compat is not a big deal.

Most the changes I see are closing parenthesis on a single line (but it's ok as it follow our "readability" principle)
and same for list and tuples with ; at the beginning of the line: having the closing parenthesis on a single line makes its even easier to cut/paste.

Julow:
Thanks for your feedback! I removed the formatting of comments and redid the change the latest version of OCamlformat.

Julow:
Here's how the formatting of comments (but not of documentation) would look: https://github.com/Julow/eliom/commit/7b05b9cfd8f2252f69b377e0c830d0c4c0ee23c3

src/lib/server/eliommod.ml
```diff
@@ -313,215 +324,248 @@ let parse_eliom_option
...
-           "Eliom: Wrong attribute value for maxdatasessionspersubnet tag"))
-  | Xml.Element ("maxpersistentsessionspergroup", [("value", v)], []) -> (
-    try
-      let i = int_of_string v in
-      set_max_persistent_sessions_per_group i
-    with Failure _ ->
-      raise
-        (Error_in_config_file
-           "Eliom: Wrong attribute value for maxpersistentsessionspergroup tag")
+    | Xml.Element ("volatiletimeout", attrs, []) ->
```
balat, line 334: Regression here (why is this indented)

src/lib/server/eliommod.ml
```diff
@@ -775,20 +842,16 @@ let gen_nothing () _ = Lwt.return Ocsigen_extensions.Ext_do_nothing
...
+    (f :
+         ?full_st_name:Eliom_common.full_state_name
+      -> ?cookie_level:[< Eliom_common.cookie_level]
+      -> recompute_expdates:bool
+      -> bool (* override configfile *)
+      -> bool (* from config file *)
+      -> Eliom_common.sitedata
+      -> float option
+      -> unit
+      ) sitedata cookie_type state_hier v =
```
balat, line 854: ?

src/lib/server/eliommod_cookies.ml
```diff
@@ -129,89 +134,87 @@ end
...
     correspond to existing (and not closed) sessions *)
-let get_cookie_info
-      now
-      sitedata
-      service_cookies
-      data_cookies
-      persistent_cookies
-      secure_cookies : 'a Eliom_common.cookie_info * 'b list
-  =
+let get_cookie_info now sitedata service_cookies data_cookies persistent_cookies
```
balat, line 137: ?

PR #882, opened 2026-06-22
client: make of_json/to_json work under wasm_of_ocaml (fix #880)

### LINKED ISSUE #880: Code audit of src/lib: 49 findings (2 high, 13 medium, 34 low)
## Code audit of `src/lib` — checklist

Tracking the findings from the audit below. Each item is a checkbox; check it when fixed (or struck through if retracted). **Methodology/caveat:** findings were produced by an automated multi-agent review and re-checked by an independent verification pass, then triaged here. They are **not** all confirmed bugs — at least one (H2) was a verifier false positive — so please confirm before acting. Line numbers refer to `master` at audit time.

**Progress:** 0 / 48 fixed · H2 retracted.

### 🔴 High

- [x] **H1 · `src/lib/eliom_client.client.ml:1316-1326`** — call_ocaml_service holds load_mutex across unwrap/do_request_data with no unlock on failure  → PR #881
  load_mutex is locked at line 1316 and unlocked only at line 1322. In between, unwrap_caml_content (line 1318) decodes/unwraps server-supplied content (can raise on malformed data) and do_request_data (line 1319) initializes client values (can raise). There is no Lwt.catch/Lwt.finalize, so any exception in these steps escapes with the mutex still held. Because load_mutex is the single global page-load mutex, this permanently blocks all later page changes and OCaml-service calls. This path is reachable for every server_function/RPC call (registered at lines 2356-2364).

- [x] ~~**H2 · `src/lib/eliom_lib.client.ml:142-148`** — of_json/to_json hardcode backend string "js_of_ocaml" and break under wasm_of_ocaml~~ — **retracted, not a bug** (wasm correctly can't use the unsafe `Json` path; see [comment](https://github.com/ocsigen/eliom/issues/880#issuecomment-4769252526))

### 🟠 Medium

- [ ] **M6 · `src/lib/client/eliommod_cookies.ml:137-151`** — Uncaught deserialization exception from localStorage poisons all cookie operations
  get_table with in_local_storage=true reads the substitute-cookie blob from localStorage and runs `intern_cookies (of_json ~typ:json_cookies (Js.to_string v))` (line 151) with no exception handling. of_json (Deriving_Json.from_string) raises if the stored JSON is malformed or has the wrong shape. localStorage is mutable, persistent, and can be corrupted by a partial write, a co-resident script, a manual edit, or a future format change. Once corrupted, every call to get_table for that host raises, and because get_table is called from update_cookie_table and get_cookies_to_send (the core read/write paths), all cookie handling for that host breaks permanently until storage is cleared, rather than degrading gracefully. The non-localStorage branch (Jstable.find ... Optdef.get) has a safe default; the localStorage branch does not protect the parse.

- [ ] **M7 · `src/lib/eliom_bus.client.ml:152-155`** — Timer-triggered flush silently drops both errors and queued messages on write failure
  In try_flush, the non-full-queue path schedules a deferred flush via `let _ = th >>= fun () -> flush t in`. The resulting thread is discarded. flush (line 143-145) first reads and CLEARS the queue, then calls t.write l. t.write (line 76-88) only swallows Failed_request 204 and reraises every other exception (network failure, server-side action error, serialization error, etc.). On the timer path that reraised exception lands in the ignored thread and is lost: no logging via `section`, no propagation into error_h (which only watches the incoming stream), and the messages in `l` are already gone from the queue, so they are never retried and never delivered. By contrast, the queue-full path (line 150) returns flush's thread to the caller of write, so the two code paths have inconsistent error semantics — the common, time-based flush silently loses data and exceptions.

- [x] **M9 · `src/lib/eliom_client.client.ml:1531-1548`** — set_template_content holds load_mutex across exception-prone code with no unlock on failure  → PR #881
  really_set locks Eliom_client_core.load_mutex at line 1537 and only unlocks it at line 1542. Between those points it calls unwrap_caml_content (line 1538) and do_request_data (line 1539). unwrap_caml_content runs Url.decode + Eliom_unwrap.unwrap on server-supplied content (eliom_client.client.ml:1287-1291) which raises synchronously on malformed/short input, and do_request_data can raise while initializing client values. Unlike set_content (lines 1726-1738) and set_content_local (lines 1613-1624), there is no surrounding Lwt.catch that calls recover()/unlocks. load_mutex is a single shared global (eliom_client_core.client.ml:266). A single failed template page load therefore leaves the mutex permanently locked, deadlocking every subsequent change_page / set_content / call_ocaml_service in the app.

- [ ] **M8 · `src/lib/eliom_comet.client.ml:77-84, 332, 340`** — "Always active"/"always idle-silent" detection broken because get_configuration self-merges the configuration list
  expected_activity (line 332) decides a page is permanently active only via the exact structural test `tbru = Some [0., 0., 0.]`, and treats `tbru = None` (line 340) as "always inactive when idle". Both rely on time_between_request_unfocused holding exactly that value. But get_configuration (lines 77-84) folds config_min over configuration_table using `first_conf configuration_table` as the initial accumulator, i.e. the very first element of the table. Hashtbl.fold then also visits that same first element, so config_min is applied to a configuration and itself. For the list field, config_min (line 62) does `Some (l1 @ l2)`, so a SINGLE configuration whose time_between_request_unfocused is `Some [0.,0.,0.]` (the result of set_always_active true) is transformed into `Some [(0.,0.,0.); (0.,0.,0.)]`. That no longer equals `Some [0.,0.,0.]`, so the always-active shortcut at line 332 never fires. Consequently a client that called set_always_active true is, after time_after_unfocus elapses, classified as `Idle instead of `Active (is_active reports the wrong state and the activity state machine drops out of active mode), directly contradicting the documented contract of set_always_active in the .mli ("if b is true, tells the client to always stay active"). The same self-duplication corrupts any non-trivial list configuration (e.g. set_time_between_requests_when_idle), doubling the entries that List.fold_left in sleep_before_next_request iterates over. The None and singleton-default cases happen to be idempotent under the `@`, which is why this hasn't been noticed for the defaults.

- [ ] **M1 · `src/lib/eliom_content_.client.ml:107-116, 143-145`** — nth (and thus insertFirstChild) returns None when the indexed child is a non-element node
  nth indexes the full childNodes list (node##.childNodes##(item n)) but then does Js.Opt.map (Dom.CoerceTo.element node) ... and returns None if that specific child fails to coerce to an element. The .mli (line 168) documents nth as "returns the nth child of [e] (first is 0)" without restriction to elements. So for an element whose nth child is a text node (e.g. whitespace from pretty-printed markup), nth returns None even though a 0th child clearly exists. This directly breaks insertFirstChild p c = appendChild ?before:(nth p 0) p c: when p's first child is a text node, nth p 0 is None, so appendChild is called with before=None, which appends c at the END of p's children instead of as the FIRST child — the exact opposite of the documented behavior.

- [ ] **M2 · `src/lib/eliom_content_.client.ml:189-217`** — nextSibling/previousSibling stop at non-element nodes, breaking insertAfter
  nextSibling/previousSibling read node##.nextSibling / node##.previousSibling (which include text/comment nodes) and then Js.Opt.bind with Dom.CoerceTo.element, returning None when the immediately adjacent node is not an element. The .mli (lines 195-202) documents these as returning "the next/previous element that has the same parent", implying non-element nodes should be skipped — instead they return None as soon as the adjacent node is text. This also corrupts insertAfter (lines 212-217): it computes before = nextSibling after; if a text node follows 'after', nextSibling returns None and appendChild ?before:None appends the element at the END of the parent rather than immediately after 'after'. Whitespace text nodes between elements are extremely common, so insertAfter mis-places nodes in realistic DOM.

- [ ] **M3 · `src/lib/eliom_parameter_base.shared.ml:445`** — add_pref_params does not recurse into TTypeFilter, so prefixed (non-localized) parameters lose their prefix when wrapped in type_checker
  In add_pref_params the TTypeFilter case returns the node unchanged: `| TTypeFilter a -> TTypeFilter a (* ??? no recursive call ??? *)`. Unlike every other tree-walking function in this file (make_suffix:203, aux:277, make_params_names:414, wrap_param_type:535, walk_parameter_tree:339), which all recurse through TTypeFilter into its inner params_type, add_pref_params stops here and never prefixes the inner parameter names.  add_pref_params is the function used by make_non_localized_parameters (line 506) to rename all inner parameters with the non-localized prefix `Eliom_common.nl_param_prefix ^ name ^ "."`. The public API allows wrapping a typed parameter in a checker via `type_checker : ('a -> unit) -> ('a,'b,'c) params_type -> ('a,'b,'c) params_type` and then passing the result to make_non_localized_parameters (its argument type `('a,[`WithoutSuffix],'b) params_type` accepts a TTypeFilter). For example `make_non_localized_parameters ~prefix ~name (type_checker check (int "x"))` yields a TNLParams whose inner TAtom keeps the bare name "x" instead of `<nl-prefix>x`. The non-localized parameter is then constructed and looked up under the wrong (unprefixed) name, so it will not be recognized as belonging to the nl-param set: get_non_localized_get/post_parameters will fail to find it and link/form construction will emit the wrong key. This is a silent correctness bug, not merely a stylistic '???' note (the matching `??? no recursive call ???` note on TConst at line 439 is harmless because TConst carries only a literal value, not a parameter name).

- [ ] **M10 · `src/lib/eliom_registration.server.ml:173`** — Malformed Cache-Control header: "max-age: N" instead of "max-age=N"
  add_cache_header emits the Cache-Control value `"max-age: " ^ string_of_int duration`, producing e.g. `Cache-Control: max-age: 3600`. The HTTP cache-control directive syntax (RFC 7234) is `max-age=<seconds>` using '=', not ':'. With a colon the directive is unrecognized, so browsers and proxies silently ignore it and do NOT cache for the requested duration. The adjacent `Some 0 -> no-cache` branch is correct, which highlights that the max-age branch is the broken one. This affects every File/File_ct/CssText/String service that passes a non-zero ~options cache duration (the .mli documents this option as the "Cache-policy: max-age header value"). It is long-standing (predates the Cohttp migration) but is a genuine functional defect: requested caching never takes effect.

- [ ] **M4 · `src/lib/eliom_request.client.ml:344-349`** — GET query parameters lost when silently redirecting to non-application content
  In the `Wrong_headers` handler of `send`, when the server returns non-application content for a GET request, the code calls `redirect_get url`. However `url` is the bare URL: the GET parameters were passed separately to `XmlHttpRequest.perform_raw_url` via `~get_args`, and `perform_raw_url` appends them internally (it does `url ^ "?" ^ Url.encode_arguments get_args`). The local `url` value never contains those arguments. Therefore `redirect_get url` navigates the browser to the page WITHOUT any of its query parameters, so a GET request such as `/service?id=42&q=foo` falls back to a full-page load of `/service` with the parameters dropped, producing a wrong/incomplete page. The parallel `expecting_process_page` success branch handles exactly this concern correctly by reconstructing the URL with `Url.add_get_args url (List.tl get_args)` (line 325), confirming that `url` here lacks the args and they must be re-appended. The `get_args` list (computed at lines 240-250) is in scope at line 345. Note the POST case is explicitly rejected at lines 346-348, so only GET is affected and GET is precisely the case that loses data.

- [ ] **M11 · `src/lib/eliom_shared.eliom:654-661`** — merge_s computes `synced` from only the last signal, not the conjunction of all
  In the server-side `React.S.Lwt.merge_s`, the fold function is `let f (acc, _acc_b) v = ... let acc_b = FakeReact.S.synced v in Lwt.return (acc, acc_b)`. It binds `_acc_b` (the running accumulator) and discards it, then sets `acc_b` to the synced status of the *current* element only. After `Lwt_list.fold_left_s f (acc, true) l`, the resulting `synced` therefore equals `FakeReact.S.synced` of the LAST element of `l` (or `true` if `l` is empty), instead of the conjunction over all elements. Compare the non-Lwt counterpart `FakeReact.S.merge` (lines 293-295), which correctly accumulates with `f (acc, acc_b) (x, b) = f acc x, acc_b && b`, and every other `*_s` function here (e.g. l2_s line 524, l6_s lines 635-638) which correctly conjoins all sources with `&&`. Consequence: if any earlier signal in the list is not synced but the last one is, `merge_s` reports `synced = true`. Per the .mli contract for `synced`, the client then assumes the server-provided initial value equals the client value and defers updates until the first client-side change — so the merged signal can render/stay at a stale value derived from an out-of-sync input until something forces a recomputation. This is a silent data-correctness bug, not just a performance hint.

- [ ] **M5 · `src/lib/eliom_state.server.ml:144-153`** — set_default_global_persistent_data_state_timeout uses wrong timeout kind (`Service instead of `Persistent)
  set_default_global_persistent_data_state_timeout calls `Eliommod_timeouts.set_default_global \`Service cookie_level override_configfile false sitedata timeout` (line 152), passing the kind `Service. The function is supposed to set the DEFAULT GLOBAL PERSISTENT state timeout (per the .mli at lines 572-576 and its name). The `kind` parameter selects which sitedata timeout slot is written: in eliommod_timeouts.ml, set_sitedata_timeout/sitedata_timeout map `Service -> sitedata.servtimeout, `Data -> data, `Persistent -> sitedata.perstimeout. So this call overwrites the service default timeout instead of the persistent one. Effect: (1) calling this function silently changes the default service-state timeout, and (2) it never changes the default persistent timeout, so subsequent get_global_persistent_data_state_timeout / persistent cookie expiry calculations keep using the old/built-in default (86400s). Compare with the correct sibling at line 113 which passes `Data for the volatile-data variant. The valid kind type is [`Service | `Data | `Persistent] (eliommod_timeouts.ml:31), so `Persistent is accepted; this is a plain typo-level defect, not a type constraint.

- [ ] **M12 · `src/lib/server/eliommod_gc.ml:318-336`** — A single failing persistent-cookie deletion permanently kills the persistent-session GC loop
  persistent_session_gc runs `let rec f () = Lwt_unix.sleep t >>= gc >>= f` via Lwt.async with no surrounding try/catch. `gc` calls Persistent_cookies.garbage_collect, which does `Lwt_list.iter_s gc_cookie cookies >>= fun _ -> Expiry_dates.remove date`. gc_cookie (lines 318-327) uses Lwt.try_bind, but only its FAILURE continuation catches exceptions (and only translates Not_found); the SUCCESS continuation `do_gc_cookie c` is returned unguarded. do_gc_cookie calls Eliommod_persess.close_persistent_state2, which performs DB/persistent-store work that can raise (transient backend error, etc.). If it raises for one cookie, Lwt_list.iter_s aborts, Expiry_dates.remove date is skipped, garbage_collect's thread fails, the failure propagates through `gc` into `f`, and because there is no catch the recursive loop terminates. The persistent-session GC then never runs again for the lifetime of the process, so expired persistent sessions/cookies accumulate without bound (resource leak) and the expiry-date index entry is never cleared.

- [ ] **M13 · `src/lib/server/eliommod_sessiongroups.ml:440-473`** — Non-atomic read-modify-write on persistent session-group table races
  Pers.add (and likewise Pers.remove at 549-567 and Pers.up at 569-580) performs a read-modify-write on the Ocsipersist group table: it `Ocsipersist.find grouptable sg` to get `(max2, cl)`, computes a new list, then `Ocsipersist.replace_if_exists grouptable sg (newmax, sess_id :: cl)`. Both ocsipersist operations yield to the Lwt scheduler, so two concurrent requests touching the same session group can both read the same `cl` and the second `replace_if_exists` overwrites the first, losing a membership update (e.g. a session never recorded in its group, or the group exceeding its `set_max` because `cut` ran on a stale list). The code itself flags this with the `(*VVV VERIFY concurrent access *)` comments (lines 417-418), and there is no transaction/lock guarding the read+write. This can corrupt group membership and defeat the per-group session limits.

### 🟡 Low

- [ ] **L1 · `src/lib/client/eliommod_cookies.ml:178-224`** — Quadratic localStorage re-serialization inside cookie folds
- [ ] **L2 · `src/lib/client/eliommod_dom.ml:601-602, 628`** — CSS text before the first @import (including @charset) is silently dropped; `charset` variable is dead
- [ ] **L3 · `src/lib/eliom_bus.server.ml:149-150`** — Dead/redundant match in stream: both branches return bus.stream
- [ ] **L4 · `src/lib/eliom_client.client.ml:594-605`** — random_int uses signed Int16Array, yielding negative/biased session ids
- [ ] **L5 · `src/lib/eliom_client_core.client.ml:166-177`** — Over-broad `with Not_found -> ()` can silently swallow client-value initialization errors
- [ ] **L6 · `src/lib/eliom_comet.client.ml:645-652, 531-551`** — Unknown comet exception silently terminates the shared service stream and all its channels
- [ ] **L7 · `src/lib/eliom_comet.server.ml:153-156`** — Stateless channel feeder thread swallows all stream exceptions silently
- [ ] **L8 · `src/lib/eliom_comet.server.ml:163-180`** — Stateless create starts consuming the stream before checking for a duplicate channel name
- [ ] **L9 · `src/lib/eliom_common.server.ml:972, 984, 998, 1007, 1087`** — Malformed client Eliom headers/params crash get_session_info (only Not_found is caught)
- [x] **L10 · `src/lib/eliom_content_.client.ml:516`** — scrollIntoView uses wrong error-context name "Css.background"  → PR #883
- [ ] **L11 · `src/lib/eliom_cscache.eliom:23-27`** — do_cache_raw unconditionally removes the cache entry on failure, can evict a newer valid value
- [ ] **L12 · `src/lib/eliom_lib_base.shared.ml:70`** — Int_map comparator uses subtraction which overflows, violating Map.OrderedType total order
- [ ] **L13 · `src/lib/eliom_mkreg.server.ml:507`** — Redundant `raise (failwith ...)` — raise is applied to a value that is never produced
- [ ] **L14 · `src/lib/eliom_parameter_base.shared.ml:772`** — TFile reconstruction error path discards the remaining GET/POST params, degrading error reporting and dropping sibling parameters
- [ ] **L15 · `src/lib/eliom_react.client.ml:54-61`** — Re-raising the exception after handling it routes it to async_exception_hook
- [ ] **L16 · `src/lib/eliom_reference.server.ml:210-216`** — Lazy initialization of persistent/global erefs has a benign-but-real read-modify-write race
- [ ] **L17 · `src/lib/eliom_registration.server.ml:688-700`** — POST argument logged unredacted when service has no name (info leak inconsistency)
- [ ] **L18 · `src/lib/eliom_request.client.ml:304-314`** — Half-XHR POST redirect ignores the redirect target URI from the header
- [ ] **L19 · `src/lib/eliom_route.client.ml:23-27, 79, 87-94`** — Client service tables grow without bound; entries are never coalesced or pruned
- [ ] **L20 · `src/lib/eliom_route.server.ml:71, 123`** — Dead parameter in find_aux
- [ ] **L21 · `src/lib/eliom_route_base.shared.ml:433`** — Misleading destructuring field names in search_by_priority_generation
- [x] **L22 · `src/lib/eliom_service.server.ml:235-236`** — Typo in attach error message ("ofservice") plus stale "attach'" name  → PR #883
- [ ] **L23 · `src/lib/eliom_service.server.ml:391-393`** — Redundant raise wrapping failwith in unregister
- [ ] **L24 · `src/lib/eliom_shared.eliom:225-231`** — client ReactiveData.RList.acc_e drops the mapped event, risking GC of the accumulation
- [ ] **L25 · `src/lib/eliom_tools.eliom:277`** — Dead/leftover code in menu (single-element-list branch of aux)
- [ ] **L26 · `src/lib/eliom_tools.eliom:462-474`** — structure_links emits rel="next" for both forward and reverse subsection links, contradicting the .mli contract
- [ ] **L27 · `src/lib/eliom_uri.shared.ml:283-293`** — #465/#471 client-app path workaround missing on the GET non-attached branch
- [ ] **L28 · `src/lib/eliom_wrap.server.ml:224-228`** — Leftover debugging code (Format.eprintf) on the lookup-failure path
- [x] **L29 · `src/lib/server/eliommod.ml:278-286, 292, 294`** — Misleading config error message: applicationscript attribute errors report 'cacheglobaldata'  → PR #883
- [ ] **L30 · `src/lib/server/eliommod.ml:519-526`** — parse_eliom_options silently swallows genuine config errors for recognized tags
- [ ] **L31 · `src/lib/server/eliommod_persess.ml:110-141`** — Browser-session persistent group capped by tab-per-session limit
- [ ] **L32 · `src/lib/server/monitor/eliom_monitor.ml:42`** — format_duration "years" segment is always emitted (wrong condition)
- [ ] **L33 · `src/lib/server/monitor/eliom_monitor.ml:32`** — fd count undercounts open file descriptors by 2
- [ ] **L34 · `src/lib/server/monitor/eliom_monitor.ml:172`** — Stylesheets loaded from third-party CDN over protocol-relative URL without SRI

hhugo:
Fixes the H2 finding reported in #880.

`to_json` and `of_json` selected their fast `Json.output` / `Json.unsafe_input` path only when `Sys.backend_type` is `Other "js_of_ocaml"`. Under `wasm_of_ocaml` the backend type is `Other "wasm_of_ocaml"`, so the branch was not taken. Two consequences on the Wasm target:

1. `of_json` called without `~typ` — which the `.mli` advertises as a valid client-side call — fell through to `None -> assert false` and crashed, whereas it works under js_of_ocaml.
2. The documented fast (un)marshaling path (see the comment above `to_json`) was silently disabled, forcing the slower `Deriving_Json` path.

### Fix

Introduce an `is_web_client` predicate matching both web backends (`js_of_ocaml | wasm_of_ocaml`). `Json.output` / `Json.unsafe_input` are provided by both runtimes, so the fast path is valid on either. The `Native | Bytecode | Other _` arm is enumerated explicitly so a future backend rename is a compile-time signal rather than a silent regression.

Built against `master` with `dune build src/lib` (OCaml 5.4.0).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/lib/eliom_lib.client.ml
+++ b/src/lib/eliom_lib.client.ml
@@ -131,21 +131,28 @@ end
 (* We do not use the deriving (un)marshaling even if typ is available
    because direct jsn (un)marshaling is very fast client side
 *)
-let to_json ?typ s =
+(* [Json.output]/[Json.unsafe_input] are provided by both the js_of_ocaml and
+   wasm_of_ocaml runtimes, so use the fast path on either web backend. *)
+let is_web_client =
   match Sys.backend_type with
-  | Other "js_of_ocaml" -> Js.to_string (Json.output s)
-  | _ -> (
+  | Other ("js_of_ocaml" | "wasm_of_ocaml") -> true
+  | Native | Bytecode | Other _ -> false
+
+let to_json ?typ s =
+  if is_web_client
+  then Js.to_string (Json.output s)
+  else
     match typ with
     | Some typ -> Deriving_Json.to_string typ s
-    | None -> Js.to_string (Json.output s))
+    | None -> Js.to_string (Json.output s)
 
 let of_json ?typ v =
-  match Sys.backend_type with
-  | Other "js_of_ocaml" -> Json.unsafe_input (Js.string v)
-  | _ -> (
+  if is_web_client
+  then Json.unsafe_input (Js.string v)
+  else
     match typ with
     | Some typ -> Deriving_Json.from_string typ v
-    | None -> assert false)
+    | None -> assert false
 
 (* Url.urlencode ~with_plus:true (Marshal.to_string x [])
     (* I encode the data because it seems that multipart does not


hhugo:
Closing this — the premise was wrong.

`Json.unsafe_input` / `Json.output` rely on js_of_ocaml's JS-native value representation. Under `wasm_of_ocaml`, OCaml values live in linear memory and that unsafe path does not apply, so the fast `Json` path genuinely cannot be used on wasm — the code must go through `Deriving_Json` (which requires a `~typ`), exactly as the original `match Sys.backend_type with Other "js_of_ocaml" -> ... | _ -> ...` already does.

The original code was correct; matching `wasm_of_ocaml` into the unsafe branch (as this PR did) would read/produce garbage on wasm. Apologies for the noise.

PR #894, opened 2026-06-30
ci(doc): cache the opam switch to skip recompiling the dependency tree

balat:
## Problem

The `Documentation` workflow takes ~8 min per push (sometimes 15+), of which **~6 min is `Install dependencies`** — recompiling the whole dependency tree (lwt, js_of_ocaml, wasm_of_ocaml, react, tyxml, ocsigenserver…) plus wodoc and odoc from source every time. `setup-ocaml@v3` only caches the base switch (the compiler), so none of that was cached.

Recent run breakdown:

| Step | Duration |
|---|---|
| Set up OCaml + Binaryen | ~25 s |
| **Install dependencies** | **5 min 58 s** |
| Build documentation (dev) | 1 min 22 s |

## Fix

Cache the whole `_opam` switch with `actions/cache`, keyed on `eliom.opam` (the dependency set) and wodoc's resolved HEAD commit. The install is split:

- **Install dependencies and wodoc** — runs only on a cache miss (deps tree + wodoc).
- **Install eliom** — runs **always**: eliom's own sources change every push, so `opam install .` re-syncs the local pin and rebuilds the package for odoc_driver to document. Only eliom recompiles on the hot path, not its dependencies.

The cache rebuilds once when `eliom.opam` or wodoc's HEAD move, then is reused.

This mirrors the pilot done on ocsigen.github.io (4 min → ~1 min). The first run on this branch is a cache miss (populates the cache); the speed-up shows on the next run.

> Note: this is the doc-CI pilot for the API-project pattern. Once confirmed, the same change rolls out to ocsigenserver, toolkit, start, tyxml, i18n, ocsipersist, lwt, reactiveData (and tuto / wodoc with their own variants).

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -56,20 +56,43 @@ jobs:
         with:
           token: ${{ github.token }}
 
-      - name: Install dependencies
+      # The bottleneck of this job is compiling the dependency tree (~6 min): lwt,
+      # js_of_ocaml, wasm_of_ocaml, react, tyxml, ocsigenserver... plus wodoc/odoc
+      # from source. setup-ocaml only caches the base switch (the compiler), so
+      # they recompiled on every push. Cache the whole _opam switch keyed on
+      # eliom.opam (the dep set) and wodoc's HEAD commit. On a cache hit the deps
+      # and wodoc are restored and only eliom itself is (re)built below; the key
+      # changes — and the deps/wodoc rebuild once — only when eliom.opam or wodoc
+      # move.
+      - name: Resolve wodoc HEAD commit
+        id: wodoc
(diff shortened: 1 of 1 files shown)


balat:
⚠️ The validation dispatch on this branch **failed, but not because of this change** — it died building `wasm_of_ocaml-compiler 6.4.0` (exit 31, `wasmoo_link_wasm.exe`: "The runtime contains unknown imports") during `opam install . --deps-only --with-doc`, which is byte-identical to the current workflow. It looks like `Aandreba/setup-binaryen@v1.0.0` now pulls a binaryen that is incompatible with wasm_of_ocaml 6.4.0's wasm runtime linking. This would break the next master doc push too, independently of caching.

So this caching PR is ready but **can't be validated until the wasm build is fixed** (binaryen pin / wasm_of_ocaml bump). Holding it. The API-project caching pattern is being validated on ocsigenserver instead.

